PR #135, opened 2024-04-16
Support W3C jwt_vc_json credential requests

vafeini:
Support W3C jwt_vc_json credential requests, first draft of pseudonym credential.

This is an example of issuing a credential in W3C VC DM 1.1 format encoded as JWT in order to test the VCI libs


babisRoutis:
@vafeini Please rebase with main

babisRoutis:
@vafeini I think that we should also update README to include the sample credential in W3C VC DM and to update the feature matrix. For instance, we support Proof where binding key is expressed as a DID using either `did:jwk` or `did:key` methods.

babisRoutis:
@vafeini  `main` branch was updated to include support for loading signing key from a keystore.
I believe that we should rebase the PR

babisRoutis:
@vafeini Please update PR description and if ready remove the `draft` marker

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/did/CreateDidUrl.kt
```diff
@@ -0,0 +1,117 @@
...
+    Curve.SECP256K1,
+    Curve.P_256,
+    Curve.P_384,
+    Curve.P_521,
+)
+
+/**
+ * did:key:MULTIBASE(base58-btc, MULTICODEC(public-key-type, raw-public-key-bytes))
+ */
+private fun createDidKey(key: JWK): DIDURL = when (key) {
```
babisRoutis, line 64: The factory methods that are currently private, have a recognizable value, on their own.

- Consider exposing them public and perhaps remove both the sealed interface `DidMethod` and the fun `createDidUrl`

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/diploma/GenerateUserDiplomaData.kt
```diff
@@ -0,0 +1,120 @@
...
+ */
+@Serializable
+data class LearningActivity(
+    @SerialName("id") val id: String,
+    @SerialName("definition") val definition: String,
+    @SerialName("startedAtTime") val started: Instant,
+    @SerialName("endedAtTime") val ended: Instant,
+)
+
+interface GenerateUserDiplomaData {
```
babisRoutis, line 117: Why call it `GenerateUserDiplomaData` and not `GetUserDiplomaData` adding the authorization context as an input parameter (take a look at PID). ?

The fact that we provide an implementation that just generates this is irrelevant, I think

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/diploma/IssueJwtVcJsonDiploma.kt
```diff
@@ -0,0 +1,188 @@
...
+        )
+
+        CredentialResponse.Issued(JsonPrimitive(jwt), notificationId)
+            .also {
+                log.info("Successfully issued diploma")
+                log.debug("Diploma data {}", it)
+            }
+    }
+
+    private fun encodeDiplomaToJwtVcJson(holderKey: JWK): String {
```
babisRoutis, line 130: Pass in the diploma and don't generate it.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/diploma/IssueJwtVcJsonDiploma.kt
```diff
@@ -0,0 +1,188 @@
...
+    override suspend fun invoke(
+        authorizationContext: AuthorizationContext,
+        request: CredentialRequest,
+        credentialIdentifier: CredentialIdentifier?,
+        expectedCNonce: CNonce,
+    ): CredentialResponse<JsonElement> = coroutineScope {
+        log.info("Handling issuance request ...")
+        val holderPubKey = async(Dispatchers.Default) { holderPubKey(request, expectedCNonce) }
+
+        val jwt = encodeDiplomaToJwtVcJson(holderPubKey.await())
```
babisRoutis, line 106: lookup the user diploma (generate it), then pass it to `encodeDiplomaToJwtVCJson(diploma, holderPubKey)`

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/diploma/IssueJwtVcJsonDiploma.kt
babisRoutis, line 130: Are we sure that holderPubKey shouldn't be a `JWK`?

It should be a DID or DIDURL, I think.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/diploma/IssueJwtVcJsonDiploma.kt
```diff
@@ -0,0 +1,188 @@
...
+
+        val credential = buildCredential {
+            credentialMetadata {
+                type(diplomaName)
+                issuer(credentialIssuerId.value.toString())
+                issueDate(_now)
+                expirationDate(_now.plusSeconds(86400))
+            }
+            credentialSubject {
+                val didUrl = createDidUrl(holderKey, DidMethod.KEY).getOrThrow()
```
babisRoutis, line 143: Whose didUrl is this? Better provide a business name.

Also given that DID and DID URL are being representing with the same class (`DIDURL`), personally I would prefer something more concrete.

Semantics seems not clear for the `holderPubKey` parameter.

babisRoutis (changes requested):

babisRoutis:
Closing this. Won't be implemented

PR #143, opened 2024-04-23
Authorization server metadata and DPoP fixes.

dzarras:
1. Introduce a new property to configure the URL to be used to fetch the metadata of the authorization server. This is not necessarily derived from the public URL of the authorization server.
2. Enable parsing of X-Forwarded headers so that the server can successfully determine the currently accessed URI when behind a reverse proxy. When this is not enabled DPoP authentication does not work because `htu` differs from `ServerHttpRequest.uri`.


babisRoutis (approved):

babisRoutis:
@dzarras Have you verified the solution using a docker image?

dzarras:
@babisRoutis Yeap. That's how I figured out we needed to enable support for the X-Forwarded headers.

PR #151, opened 2024-04-30
Add the ability to load a key-pair from a keystore, for SD-JWT signing.

### LINKED ISSUE #147: Load signing key from a keystore
Currently, application generates a random key pair, for signing credentials with format SD-JWT-VC during [startup](https://github.com/eu-digital-identity-wallet/eudi-srv-pid-issuer/blob/35997a9f578ec3194606406230e8dae0b240b4d3/src/main/kotlin/eu/europa/ec/eudi/pidissuer/PidIssuerApplication.kt#L340).

We could load this key pair from an externally defined keystore, similar to that the [verifier](https://github.com/eu-digital-identity-wallet/eudi-srv-web-verifier-endpoint-23220-4-kt/blob/1ba6ccc186fb2f1b8c9beeb119142def1c71121e/src/main/kotlin/eu/europa/ec/eudi/verifier/endpoint/VerifierContext.kt#L208) does.

dzarras:
Closes #147


README.md
```diff
@@ -147,6 +147,35 @@ Variable: `ISSUER_PID_SD_JWT_VC_DEFERRED`
 Description: Whether PID issueance in *SD JWT VC* format should be *deferred* or *immediate*  
 Default value: `false` (i.e. immediate issuance)
 
+Variable: `ISSUER_PID_SD_JWT_VC_SIGNING_KEY`
```
babisRoutis, line 150: I don't think that we need to define so many keys (per credential, per format etc).

I single key pair would be enough.

babisRoutis (changes requested):
Do we really need multiple singing keys?
I don't think so

README.md
dzarras: Could the same key be used both for sd-jwt and mdoc signing? 
If both require an ECKey, then we can proceed as you suggest.

babisRoutis (approved):

PR #152, opened 2024-05-01
Change mso_mdoc signing implementation

### LINKED ISSUE #153: Use a single signing key for all formats, owned by the isser.
Currently, issuer depends on an external service for signing credentials in `mso_mdoc` format.
In addition, this external service owns the key that is being used to sign the credentials.

For SD-JWT-VC and W3C DM formats, it is the issuer that controls the signing key (either randomly generated upon application startup or externally provided via a keystore).

This is problematic. Issuer should use a single signing key used for all formats to simplify key management

babisRoutis:
Closes #153 

This PR removes the dependency to the external service and implements `mso_mdoc` signing internally.
Now, issuer is able to use a single signing key (loaded by external key store) for all formats.

- [x] Migrate PID in `mso_mdoc` to the internal signing service
- [x] Migrate mDL in `mso_mdoc` to the internal signing service
- [x] Remove dependency to the external signing service
- [x] Load a single signing key from external keystore (Perhaps remove support for randomly generated keys)
- [x] Possibly, add a section in the README, providing details on how to create the keystore


dzarras:
@babisRoutis could you please check and let me know of any comments?

PR #203, opened 2024-08-23
Fix issues with Credential Response Encryption.

### LINKED ISSUE #204: Issues with Credential Response Encryption
1. ~~RequestedResponseEncryption.Required does not accept JWKs with keyUse null. When keyUse is null, the JWK can be used for all purposes, including encryption.~~
2. During issuance when CredentialRequestTO contains no CredentialResponseEncryptionTO, the issuance continues and an unencrypted credential is issued even when the issuer is configured to require encryption.

dzarras:
1. ~~RequestedResponseEncryption.Required now accepts JWK with keyUse null as well. When keyUse is null, the JWK can be used for all purposes, including encryption.~~
2. Properly verify during issuance that RequestedResponseEncryption is supported by the issuer's configured CredentialResponseEncryption. Previously when CredentialResponseEncryption was set to Required, but the CredentialRequestTO contained no CredentialResponseEncryptionTO, the issuance would continue and an unencrypted credential would be issued, even though the issuer was configured to require encryption.

Closes #204

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/IssueCredential.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/IssueCredential.kt
@@ -257,6 +257,7 @@ class IssueCredential(
                 when (unresolvedRequest) {
                     is UnresolvedCredentialRequest.ByFormat ->
                         unresolvedRequest.credentialRequest to null
+
                     is UnresolvedCredentialRequest.ByCredentialIdentifier ->
                         resolve(unresolvedRequest) to unresolvedRequest.credentialIdentifier
                 }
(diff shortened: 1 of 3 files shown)


src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/IssueCredential.kt
```diff
@@ -459,54 +460,63 @@ private fun ProofTo.toDomain(): UnvalidatedProof = when (type) {
...
             }
-        }
 
-        RequestedResponseEncryption.Required(Json.encodeToString(key), algorithm, method)
-            .bind()
-            .also { it.ensureIsSupported() }
+            is CredentialResponseEncryption.Required -> {
+                if (this !is RequestedResponseEncryption.Required) {
+                    // credential response encryption required by issuer but not required by client
+                    throw IllegalArgumentException("credential response encryption is required")
```
babisRoutis, line 493: Please use error( .. ) instead of IllegalArgumentException

Further more, why use if ... throw and not the ensure()

PR #211, opened 2024-09-09
Remove dependency to common-did lib

### LINKED ISSUE #208: Improve dependency security and use less unknown third-party registrys (jitpack-usage, danubtech)
Hi, We are evaluating the projects for further usage and would like to propose some security improvements in the build process.

**Problem:** 

A widely used approach today for many organizations is to only allow dependency usage through internal registries, so that the sources of trust are more evaluated, can be scanned, and provide a wide overview of what is used in the organization.

This project uses a few different third-party registries that do not have strict security policies. To evaluate the project, one would have to trust all of these in the organization's repository.

Registries in use are:
Maven Central, Gradle Plugin Portal, Waltid registry, Danubtech registry, JitPack.

This means that for someone to build this project, they would have to allow all five registries and trust them in their security evaluation. Now, Maven Central and Gradle Plugin Portal are fine - they have basic security verification (https://central.sonatype.org/faq/central-security/ and https://blog.gradle.org/plugin-portal-validations) of who the submitter is, etc., and are widely used, with published security policies. They are de-facto registries in the ecosystem.

However, the other three add more work for the organization and open up the usage of unknown third-party registries.

As for Waltid, it looks trustworthy based on evaluation and business area.

But then it starts to get looser. Who is the organization hosting Danubtech, and what are their safety and security policies? JitPack publishing is even looser - anyone can publish anything to JitPack at any time. I would be reluctant to allow that in an internal registry.

**Solution Proposal**: 

Ease the building of the project and improve its security and trust. Work on removing the added third-party registries that are not hosted on the de-facto registry platforms (that is, Maven Central, Gradle Plugin Portal) with clear security policies. That means starting with Danubtech and JitPack. The most important one would be the JitPack dependency, as there is no security validation or verification for anything published on JitPack.

babisRoutis:
Closes #208

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -59,9 +59,6 @@ dependencies {
     implementation(libs.zxing) {
         because("To generate a QR Code for Credentials Offer URI")
     }
-    implementation(libs.did.common) {
-        because("To support parsing of DID URLs")
-    }
     implementation(libs.multiformat) {
         because("To support resolution of did:key")
     }
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -17,7 +17,7 @@ dependencyCheck = "10.0.3"
 sonarqube = "5.0.0.4638"
 bootstrap = "5.3.3"
 jacoco = "0.8.11"
-didCommon = "1.13.0"
+
 multiformat = "1.1.0"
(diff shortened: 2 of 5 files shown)


dzarras (approved):
Looks good to me

PR #216, opened 2024-09-25
Draft14 - Batch Issuance

### LINKED ISSUE #206: Update metadata to draft 14
- [x] Remove batch endpoint
- [x] Add new `batch_credential_issuance`

### LINKED ISSUE #207: Credential endpoint to support issuing multiple instance
As per draft 14 of OpenId4VCI, Credential Endpoint can support to support issuing multiple instance of a `credential_configuration_id`

babisRoutis:
Closes #207 
Closes #206

To do:
- [x]  Test it with vci lib draft 13
- [x] Test it with vci lib draft 14 (https://github.com/eu-digital-identity-wallet/eudi-lib-jvm-openid4vci-kt/pull/312)
- [x] Consider including to the PR #206. If a limit is set for the number of the proofs this should be validated, while processing the issuance request 

The new version of the PID Issuer should be compatible with d13 wallets, except the batch endpoint.
Wallets supporting d14, should be able to use the new feature of request batch issuance via the usual credential endpoint.

> [!NOTE]
> In Draft 14, `notification_id` & the response of deferred endpoint were not properly adjusted against the `credentials` array. 
> This PR adapts the changes that will be introduced in draft 15, for the deferred response.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ the [EUDI Wallet Reference Implementation project description](https://github.co
 
 ## Overview
 
-An implementation of a credential issuing service, according to OpenId4VCI - draft13
+An implementation of a credential issuing service, according to OpenId4VCI - draft14
 
 The service provides generic support for `mso_mdoc` and `SD-JWT-VC` formats using PID and mDL as an example
 and requires the use of a suitable OAUTH2 server.
--- a/README.md
+++ b/README.md
@@ -28,19 +28,18 @@ and requires the use of a suitable OAUTH2 server.
 
 ### OpenId4VCI coverage
 
-| Feature                                                   | Coverage                                                  |
-|-----------------------------------------------------------|-----------------------------------------------------------|
-| Authorization Code flow                                   | ✅ Using a suitable OAUTH2 server                          |
-| Pre-authorized code flow                                  | ❌                                                         |
-| mso_mdoc format                                           | ✅                                                         |
-| SD-JWT-VC format                                          | ✅ Except revocation list & meta                           |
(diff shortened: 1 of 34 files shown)


babisRoutis:
@dzarras 


Draft 14, introduces the top-level metadata attribute `batch_credential_issuance`. If present it must include attribute `batch_size`.

ISO profiles, define a very similar attribute for `mso_mdoc` format named `policy`. If present includes the optional attribute `batch_size`.

Leaving aside the granularity difference (VCI attribute is global to the issuer, ISO attribute is per `credential_configuration_id`), I think that issuer should be consistent. That is 

If a value has been set to `batch_credential_issuance/batch_size`, populate the `mso_mdoc`-specific `policy/batch_size` with the same value.

Currently, although we set a value to the global parameter it is not propagated to the pid & mDL in `mso_mdoc`

babisRoutis:
@dzarras 

Unrelated to the PR, can you also check `MetaDataApi`

It seems that the route bellow is not being used. Certainly points to an method which has not been implemented.
```kotlin
 GET(WELL_KNOWN_JWKS, accept(MediaType.APPLICATION_JSON)) { _ ->
            handleGetJwtIssuerJwkSet()
}
```

dzarras:
@babisRoutis Can you check again? The latest commit provides support for setttin `batch_credential_issuance/batch_size` to the mso_mdoc-specific `policy/batch_size` with the same value, for each configured `MsoMdocCredentialConfiguration`.

babisRoutis:
@dzarras BTW, before merging this, perhaps we should release, whatever we have in main (to fix d13 support) and then release.

Issue #232, opened 2024-11-04
Nonce Endpoint


dzarras:
Quoting various parts of the latest Editor's Draft from [OpenId 4 VCI](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html).

A Credential Issuer that requires `c_nonce` values to be incorporated into proofs in the Credential Request 
MUST offer a Nonce Endpoint using the `nonce_endpoint` Credential Issuer Metadata parameter.

A request for a nonce is made by sending an HTTP POST request to the URL provided in the `nonce_endpoint` 
Credential Issuer Metadata parameter. e.g.:
```
POST /nonce HTTP/1.1
Host: credential-issuer.example.com
Content-Length: 0
```

The Credential Issuer provides a nonce value in the HTTP response with a 2xx status code in the message body of the
HTTP response using the `application/json` media. e.g.:
```
HTTP/1.1 200 OK
Content-Type: application/json
Cache-Control: no-store

{
  "c_nonce": "wKI4LT17ac15ES9bw8ac4",
  "c_nonce_expires_in": 120
}
```

The proof(s) in the `proof` or `proofs` parameter of a Credential Request MUST incorporate the Credential Issuer
Identifier (audience), and optionally a `c_nonce` value generated by the Credential Issuer to allow the Credential
Issuer to detect replay.
The initial `c_nonce` value can be returned in a Nonce Response, or in a Credential Error Response.
To validate a key proof, the Credential Issuer MUST ensure among others that if the server had previously provided a
`c_nonce`, the nonce in the key proof matches the server-provided `c_nonce` value.

Credential Response can be immediate or deferred and can contain one or more Credentials with the same Credential
Configuration and Credential Dataset depending on the Credential Request. Credential response MAY contain a `c_nonce`
parameter containing a nonce to be used to create a proof of possession of key material when requesting a 
Credential. When received, the Wallet MUST use this nonce value for its subsequent Credential Requests until the 
Credential Issuer provides a fresh nonce.

For errors related to the Credential Request's proof of possession, the error code `invalid_proof` is to be used. This
error code indicates the `proof` or `proofs` parameter in the Credential Request is invalid: 
(1) if both fields are missing, or (2) both are present simultaneously, or (3) one of the provided key proofs is 
invalid, or (4) if a `c_nonce` was previously provided and at least one of the key proofs is not linked to a valid 
`c_nonce` value.

[...]

PR #235, opened 2024-11-12
Implement self-authenticating CNonce values using Encrypted JWTs.

dzarras:
Implements self-authenticating CNonce using EncryptedJWTs. Implementation is based on ideas presented in https://datatracker.ietf.org/doc/html/draft-demarco-nonce-endpoint

- CNonce is no longer bound to an Access Token
- CNonce is encoded as an Encrypted JWT that can be decrypted only by its Issuer
- CNonce no longer needs to be persisted

This PR also removes any functionality related to CNonce persistence since it's no longer needed.
Additionally this PR reworks Proof validation a bit, to reduce code duplication, and improve code readability.

This PR **DOES NOT** implement the [Nonce Endpoint](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#name-nonce-endpoint) as described in OpenID 4 VCI Draft 15. It merely adds some building blocks that could be used for its implementation.

Relates to #232


babisRoutis:
@dzarras Why signing?
Is there any other party to verify the signature? I guess not.
To the wallet, `c_nonce` is opaque.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/DecryptCNonceWithNimbus.kt
```diff
@@ -0,0 +1,95 @@
...
+            )
+        }
+
+    override suspend fun invoke(encrypted: String): Result<CNonce> = result {
+        val jwt = EncryptedJWT.parse(encrypted)
+        val claimSet = processor.process(jwt, null)
+        CNonce(
+            requireNotNull(claimSet.getStringClaim("cnonce")) { "missing 'cnonce' claim" },
+            requireNotNull(claimSet.issueTime.toInstant()) { "missing 'iat' claim" },
+            requireNotNull(claimSet.getLongClaim("exi")?.let { Duration.ofSeconds(it) }) { "missing 'exi' claim" },
```
babisRoutis, line 92: Given that there is an `exp` claim, what is the use of `exi`?

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/DecryptCNonceWithNimbus.kt
```diff
@@ -0,0 +1,95 @@
...
+import org.bouncycastle.jce.provider.BouncyCastleProvider
+import java.time.Duration
+
+/**
+ * Nimbus implementation of [DecryptCNonce] that decrypts an [EncryptedJWT] as a [CNonce].
+ */
+internal class DecryptCNonceWithNimbus(
+    private val issuer: CredentialIssuerId,
+    private val decryptionKey: RSAKey,
+    private val signingKey: IssuerSigningKey,
```
babisRoutis, line 46: We don't need signing. There is not another party to verify it.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/EncryptCNonceWithNimbus.kt
```diff
@@ -0,0 +1,86 @@
...
+import eu.europa.ec.eudi.pidissuer.port.out.jose.EncryptCNonce
+import org.bouncycastle.jce.provider.BouncyCastleProvider
+import java.util.*
+
+/**
+ * Nimbus implementation of [EncryptCNonce] that encrypts [CNonce] as an [EncryptedJWT].
+ */
+internal class EncryptCNonceWithNimbus(
+    private val issuer: CredentialIssuerId,
+    private val signingKey: IssuerSigningKey,
```
babisRoutis, line 44: No need for signing

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/EncryptCNonceWithNimbus.kt
```diff
@@ -0,0 +1,86 @@
...
+        val signedJwt = SignedJWT(
+            JWSHeader.Builder(signingKey.signingAlgorithm)
+                .type(JOSEObjectType("cnonce+jwt"))
+                .build(),
+            JWTClaimsSet.Builder()
+                .apply {
+                    issuer(issuer.externalForm)
+                    audience(issuer.externalForm)
+                    claim("cnonce", cnonce.nonce)
+                    claim("exi", cnonce.expiresIn.seconds)
```
babisRoutis, line 65: Please remove `exi`. There is no need for two claims to track expiration

dzarras:
> @dzarras Why signing? Is there any other party to verify the signature? I guess not. To the wallet, `c_nonce` is opaque.

No reason at all. Just wanted to test out how to create a Signed+Encrypted JWT with Nimbus.
We can easily revert this, it's a single commit.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/ValidateJwtProof.kt
```diff
@@ -163,9 +171,7 @@ private fun processor(
             jwtClaimsSetVerifier =
                 DefaultJWTClaimsVerifier<SecurityContext?>(
                     credentialIssuerId.externalForm, // aud
-                    JWTClaimsSet.Builder()
-                        .claim("nonce", expected.nonce)
-                        .build(),
+                    JWTClaimsSet.Builder().build(),
                     setOf("iat", "nonce"),
```
babisRoutis, line 175: Let's remove `nonce`. There is no need to hard-code requirements for nonce at this level

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/ValidateJwtProof.kt
```diff
@@ -37,43 +37,52 @@ import com.nimbusds.jwt.proc.DefaultJWTProcessor
...
+        }
+        check(proofType is ProofType.Jwt)
+
+        return validate(unvalidatedProof, proofType)
     }
-    check(proofType is ProofType.Jwt)
 
-    return validateJwtProof(credentialIssuerId, unvalidatedProof, expectedCNonce, proofType)
+    context (Raise<IssueCredentialError.InvalidProof>)
+    private suspend fun validate(
```
babisRoutis, line 69: Let's extract a new method that returns the pair of `CredentialKey` and the value of the `nonce` claim, if present.

```kotlin

private suspend fun Raise<IssueCredentialError.InvalidProof>.credentialKeyNonce(
    unvalidatedProof: UnvalidatedProof.Jwt,
    proofType: ProofType.Jwt,
): Pair<CredentialKey, String?> = TODO()
```

Then we can refactor validate() to enforce (or not) the presence of a valid `c_nonce`

To do so, we need to pass an optional decrypt

```kotlin

private suspend fun Raise<IssueCredentialError.InvalidProof>.validate(
    unvalidatedProof: UnvalidatedProof.Jwt,
    proofType: ProofType.Jwt,
    decryptCNonce: DecryptCNonce?
): Pair<CredentialKey, CNonce?> {
    
    
    val (key, nonce) = credentialKeyNonce(unvalidatedProof, proofType)
    if (decryptCNonce != null) {
        // require nonce not null and valie
        }
    
}

```

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/CNonce.kt
```diff
@@ -19,7 +19,6 @@ import java.time.Duration
 import java.time.Instant
 
 data class CNonce(
-    val accessToken: String,
     val nonce: String,
     val activatedAt: Instant,
     val expiresIn: Duration,
```
babisRoutis, line 24: I don't the CNonce needs to have "shape".

CNonce = Nonce

Only two classes should be aware of the structure of the c_nonce
Generator / Validator.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/IssueCredential.kt
```diff
@@ -302,12 +301,10 @@ private val log = LoggerFactory.getLogger(IssueCredential::class.java)
  * Usecase for issuing a Credential.
  */
 class IssueCredential(
```
babisRoutis, line 303: I don't like the removal of clock.

We are relying too much on the dependency injection. 
Anyways, don't change it.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/IssueCredential.kt
```diff
@@ -302,12 +301,10 @@ private val log = LoggerFactory.getLogger(IssueCredential::class.java)
...
-    private val clock: Clock,
     private val credentialIssuerMetadata: CredentialIssuerMetaData,
     private val resolveCredentialRequestByCredentialIdentifier: ResolveCredentialRequestByCredentialIdentifier,
-    private val loadCNonceByAccessToken: LoadCNonceByAccessToken,
-    private val genCNonce: GenerateCNonce,
-    private val upsertCNonce: UpsertCNonce,
+    private val generateCNonce: GenerateCNonce,
```
babisRoutis, line 306: I was expecting a single `generateCNonce` that hides the how this is done.

`IssueCredential` doens't have to know how `c_nonce` is being generated

babisRoutis (changes requested):

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/DecryptCNonceWithNimbus.kt
dzarras: `exi` is the IANA registered claim for `Expires In`.
For CNonce we need `iat`, `cnonce` and `exi`. `exp` is basically `iat` + `exi`.

https://www.iana.org/assignments/jwt/jwt.xhtml

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/EncryptCNonceWithNimbus.kt
dzarras: I'd prefer to remove `exp` and keep `exi` because it better maps to our `CNonce` structure.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/IssueCredential.kt
dzarras: Would something like this make sense?


```kotlin
fun interface GenerateCNonce {
  //Return the `serialized encrypted jwt` cnonce and `expires in`
  suspend operator fun invoke(): Pair<String, Duration>
}

fun interface VerifyCNonce {
  //Decrypts jwt, verifies cnonce is not expires, and returns the cnonce value
  suspend operator fun invoke(value: String): Result<Nonce>
}
```

`Nonce` as defined in Nimbus.

dzarras: If we follow this approach, we can remove the `CNonce` data class as well.

babisRoutis (approved):

PR #242, opened 2024-11-19
Removed deprecated context receivers

### LINKED ISSUE #223: Remove usage of Kotlin's Context Receivers
Context Receivers are being phased out. Their usage in pid-issuer must be removed.
We need to define a way forward and start implementing the required changes.

Reference: https://kotlinlang.org/docs/whatsnew2020.html#phased-replacement-of-context-receivers-with-context-parameters

babisRoutis:
Closes #223

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/JWKExtensions.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/JWKExtensions.kt
@@ -19,11 +19,16 @@ import arrow.core.raise.Raise
 import com.nimbusds.jose.jwk.ECKey
 import com.nimbusds.jose.jwk.JWK
 
-/**
- * Converts this [JWK] to an [ECKey] or raises [onFailure].
- */
-internal fun <E> Raise<E>.toECKeyOrFail(jwk: JWK, onFailure: () -> E): ECKey =
-    when (jwk) {
-        is ECKey -> jwk
-        else -> raise(onFailure())
-    }
+internal interface JWKExtensions<in Error> : Raise<Error> {
+    /**
+     * Converts this [JWK] to an [ECKey] or raises [onFailure].
+     */
+    fun JWK.toECKeyOrFail(onFailure: () -> Error): ECKey =
+        when (this) {
+            is ECKey -> this
+            else -> raise(onFailure())
+        }
+}
+
+internal fun <Error> Raise<Error>.jwkExtensions(): JWKExtensions<Error> =
+    object : JWKExtensions<Error>, Raise<Error> by this {}
--- a/src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/JWKExtensions.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/JWKExtensions.kt
@@ -22,9 +22,8 @@ import com.nimbusds.jose.jwk.JWK
 /**
  * Converts this [JWK] to an [ECKey] or raises [onFailure].
  */
-context(Raise<E>)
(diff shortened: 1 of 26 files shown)


babisRoutis:
@dzarras  low priority, but please take a look

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/ValidateProofs.kt
```diff
@@ -36,34 +39,32 @@ internal class ValidateProofs(
...
-                is UnvalidatedProof.Jwt -> async {
-                    withContext(Dispatchers.Default) {
-                        validateJwtProof(it, credentialConfiguration)
-                    }
+    ): Either<InvalidProof, NonEmptyList<JWK>> = coroutineScope {
+        either {
+            val credentialKeysAndCNonces = unvalidatedProofs.map {
+                when (it) {
+                    is UnvalidatedProof.Jwt ->
+                        withContext(Dispatchers.Default) { validateJwtProof(it, credentialConfiguration).bind() }
```
dzarras, line 51: @babisRoutis If we remove `async` we might as well remove `withContext` and `coroutineScope` as well.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/JWKExtensions.kt
```diff
@@ -22,9 +22,8 @@ import com.nimbusds.jose.jwk.JWK
 /**
  * Converts this [JWK] to an [ECKey] or raises [onFailure].
  */
-context(Raise<E>)
-internal fun <E> JWK.toECKeyOrFail(onFailure: () -> E): ECKey =
-    when (this) {
-        is ECKey -> this
+internal fun <E> Raise<E>.toECKeyOrFail(jwk: JWK, onFailure: () -> E): ECKey =
```
dzarras, line 25: @babisRoutis I think the method would feel more 'organic' if it was defined as `JWK.toECKeyOrFail(onFailure: () -> E): Either<E, ECKey>`

dzarras:
In many instances methods that were defined as:

```kotlin
context(Raise<E>)
fun Receiver.f(param1: P1, param2: P2): Result { }
```

have been rewritten as

```kotlin
fun Raise<E>.f(receiver: Receiver, param1: P1, param2: P2): Result { }
```

I feel like such methods should have been rewritten instead as:

```kotlin
fun Receiver.f(param1: P1, param2: P2): Either<E, Result> { }
```

For me `f()` remains an extension function of `Receiver`. What changes, is how we communicate the errors we produce.

@babisRoutis What do you think? I feel like this approach would produce more comprehensive and readable code.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/JWKExtensions.kt
babisRoutis: Reworked `JwkExtensions` as follows:

Introduced a new interface 
```korlin
internal interface JWKExtensions<in Error> : Raise<Error> {
    /**
     * Converts this [JWK] to an [ECKey] or raises [onFailure].
     */
    fun JWK.toECKeyOrFail(onFailure: () -> Error): ECKey =
        when (this) {
            is ECKey -> this
            else -> raise(onFailure())
        }
}
```
This interface has the `toECKeyOrFail` method having the `JWK` as receiver parameter, which  seems more natural choice.
In addition, a new method was added that "lifts" the above interface/method in the context of a `Raise<E>` 

```kotlin
internal fun <Error> Raise<Error>.jwkExtensions(): JWKExtensions<Error> =
    object : JWKExtensions<Error>, Raise<Error> by this {}
```

dzarras (approved):

Issue #275, opened 2025-02-13
Update Defered Endpoint


babisRoutis:
Response 
======

- [x] Remove `credential`
- [x] Update the elements of `credentials` to be JSON object with `credential` claim

Issue #273, opened 2025-02-13
Update Credential Endpoint


babisRoutis:
Request
=======
- [ ]  Replace `format` with `credential_configuration_id`

Response
========
- [ ] Remove `c_nonce`, `c_nonce_expires_in`
- [ ] Remove `credential`
- [ ] Update the elements of `credentials` to be JSON object with `credential` claim

PR #283, opened 2025-02-14
Credential Endpoint/Deferred Credential Endpoint draft15 alignment

dzarras:
See:
- #273 
- #275

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/GetDeferredCredential.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/GetDeferredCredential.kt
@@ -44,50 +44,40 @@ sealed interface DeferredCredentialSuccessResponse {
      */
     @Serializable
     data class PlainTO(
-        val credential: JsonElement?,
-        val credentials: JsonArray?,
+        val credentials: List<CredentialTO>,
         @SerialName("notification_id") val notificationId: String? = null,
     ) : DeferredCredentialSuccessResponse {
         init {
-            require((credential != null) xor (credentials != null)) {
-                "exactly one of 'credential' or 'credentials' must be provided"
-            }
-            credential?.also { credential ->
-                require(credential is JsonObject || (credential is JsonPrimitive && credential.isString)) {
-                    "credential must either be a JsonObject or a string JsonPrimitive"
-                }
-            }
-            credentials?.forEach { credential ->
-                require(credential is JsonObject || (credential is JsonPrimitive && credential.isString)) {
-                    "credentials must contain either JsonObjects or string JsonPrimitives"
-                }
+            require(credentials.isNotEmpty()) {
+                "credentials must not be empty"
             }
         }
 
         companion object {
             /**
-             * Single credential has been issued.
-             */
-            fun single(
-                credential: JsonElement,
-                notificationId: String?,
-            ): PlainTO = PlainTO(
-                credential = credential,
-                credentials = null,
-                notificationId = notificationId,
-            )
-
-            /**
              * Multiple credentials have been issued.
              */
-            fun multiple(
+            operator fun invoke(
                 credentials: JsonArray,
                 notificationId: String?,
             ): PlainTO = PlainTO(
-                credential = null,
-                credentials = credentials,
+                credentials = credentials.map { CredentialTO(it) },
                 notificationId = notificationId,
             )
+        }
+
+        /**
+         * A single issued Credential.
+         */
+        @Serializable
+        data class CredentialTO(
+            @Required val credential: JsonElement,
+        ) {
+            init {
+                require(credential is JsonObject || (credential is JsonPrimitive && credential.isString)) {
+                    "credential must be either a JsonObjects or a string JsonPrimitive"
+                }
+            }
         }
     }
 
@@ -132,17 +122,10 @@ class GetDeferredCredential(
             is LoadDeferredCredentialResult.IssuancePending -> raise(GetDeferredCredentialErrorTO.IssuancePending)
             is LoadDeferredCredentialResult.InvalidTransactionId -> raise(GetDeferredCredentialErrorTO.InvalidTransactionId)
             is LoadDeferredCredentialResult.Found -> {
-                val plain = when (loadDeferredCredentialResult.credential.credentials.size) {
-                    1 -> DeferredCredentialSuccessResponse.PlainTO.single(
-                        loadDeferredCredentialResult.credential.credentials.head,
-                        loadDeferredCredentialResult.credential.notificationId?.value,
-                    )
-
-                    else -> DeferredCredentialSuccessResponse.PlainTO.multiple(
-                        JsonArray(loadDeferredCredentialResult.credential.credentials),
-                        loadDeferredCredentialResult.credential.notificationId?.value,
-                    )
-                }
+                val plain = DeferredCredentialSuccessResponse.PlainTO(
+                    JsonArray(loadDeferredCredentialResult.credential.credentials),
+                    loadDeferredCredentialResult.credential.notificationId?.value,
+                )
 
                 when (loadDeferredCredentialResult.responseEncryption) {
                     RequestedResponseEncryption.NotRequired -> plain
--- a/src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/GetDeferredCredential.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/GetDeferredCredential.kt
@@ -26,10 +26,7 @@ import eu.europa.ec.eudi.pidissuer.port.out.persistence.LoadDeferredCredentialRe
 import kotlinx.serialization.Required
 import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
-import kotlinx.serialization.json.JsonArray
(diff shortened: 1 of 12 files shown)


src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/GetDeferredCredential.kt
```diff
@@ -44,51 +44,41 @@ sealed interface DeferredCredentialSuccessResponse {
...
                 notificationId = notificationId,
             )
         }
+
+        /**
+         * A single issued Credential.
+         */
+        @Serializable
+        data class CredentialTO(
+            @Required val credential: JsonElement,
```
babisRoutis, line 74: Each credential may have additional JSON elements. 
This is not reflected to `CredentialTO`

So, `CredentialTO` should be serialized into a JsonObject having a mandatory `credential` claim and optionally, additional claims.

Two possible implementations that I can think of:

- Make `CredentialTO` a value class containing a JsonObject. Factory method will enforce the presence of `credential` claim.
- Keep `CredentialTO` as a data class, add another optional `additionalInfo: JsonObject?` and provide a custom serializer that would merge `credential` & `additionalInfo` into a JsonObject

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/IssueCredential.kt
```diff
@@ -60,12 +51,14 @@ data class ProofTo(
...
     @SerialName("ldp_vp")
     val ldpVp: String? = null,
+    val attestation: String? = null,
 )
 
 @Serializable
 data class ProofsTO(
     @SerialName("jwt") val jwtProofs: List<String>? = null,
     @SerialName("ldp_vp") val ldpVpProofs: List<String>? = null,
+    @SerialName("attestation") val attestations: List<String>? = null,
```
babisRoutis, line 61: I have the impression that additional checks apply for `attestation` case.
If provided, it should contain a single element.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/IssueCredential.kt
```diff
@@ -75,22 +68,16 @@ data class CredentialResponseEncryptionTO(
...
 @Serializable
 data class CredentialRequestTO(
-    val format: FormatTO? = null,
-    val proof: ProofTo? = null,
-    val proofs: ProofsTO? = null,
     @SerialName("credential_identifier")
     val credentialIdentifier: String? = null,
+    @SerialName("credential_configuration_id")
+    val credentialConfigurationId: String? = null,
+    val proof: ProofTo? = null,
```
babisRoutis, line 77: Mark `proof` as deprecated. It will be removed from a subsequent draft

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/IssueCredential.kt
```diff
@@ -194,68 +165,52 @@ sealed interface IssueCredentialResponse {
...
-            ): PlainTO = PlainTO(transactionId = transactionId, nonce = nonce)
+            fun deferred(transactionId: String): PlainTO = PlainTO(transactionId = transactionId)
+        }
+
+        /**
+         * A single issued Credential.
+         */
+        @Serializable
+        data class CredentialTO(
+            @Required val credential: JsonElement,
```
babisRoutis, line 207: Please check my comments on `CredentialTO` above

babisRoutis (changes requested):

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/IssueCredential.kt
dzarras: My bad.

From the spec:
> attestation: A JWT [[RFC7519](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#RFC7519)] representing a key attestation without using a proof of possession of the cryptographic key material that is being attested. When a proof_type parameter in a proof object is set to attestation, the object MUST also contain an attestation parameter that includes a JWT as defined in [Section 8.2.1.3](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#attestation-proof-type).

My understanding is that it is a single JWT string. Not a list with a single JWT string.

babisRoutis: For draft 15, use the following as a guide

```json
{
  "proof_type": "attestation",
  "attestation": "<key attestation in JWT format>"
}
``` 

After d15, it will change again (https://github.com/openid/OpenID4VCI/pull/453)

dzarras: Disregard my previous comment. This is the description for the `proof` property of Credential Request.

Quoting section [8.2.1 Proof Types](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-proof-types):
> [8.2.1. ](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#section-8.2.1)[Proof Types](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-proof-types)
This specification defines the following proof types:
`jwt`: A JWT [[RFC7519](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#RFC7519)] is used for proof of possession. When a proof_type parameter in a proof object is set to jwt, it MUST also contain a jwt parameter that includes a JWT as defined in [Section 8.2.1.1](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#jwt-proof-type). When a proofs object is using a jwt proof type, it MUST include a jwt parameter with its value being an array of JWTs, where each JWT is formed as defined in [Section 8.2.1.1](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#jwt-proof-type).
`ldp_vp`: A W3C Verifiable Presentation object signed using the Data Integrity Proof [[VC_Data_Integrity](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#VC_Data_Integrity)] as defined in [[VC_DATA_2.0](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#VC_DATA_2.0)] or [[VC_DATA](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#VC_DATA)] is used for proof of possession. When a proof_type parameter in a proof object is set to ldp_vp, it MUST also contain an ldp_vp parameter that includes a [W3C Verifiable Presentation](https://www.w3.org/TR/vc-data-model-2.0/#presentations-0) defined in [Section 8.2.1.2](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#ldp-vp-proof-type). When a proofs object is using a ldp_vp proof type, it MUST include an ldp_vp parameter with its value being an array of [W3C Verifiable Presentations](https://www.w3.org/TR/vc-data-model-2.0/#presentations-0), where each of these W3C Verifiable Presentation is formed as defined in [Section 8.2.1.2](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#ldp-vp-proof-type).
`attestation`: A JWT [[RFC7519](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#RFC7519)] representing a key attestation without using a proof of possession of the cryptographic key material that is being attested. When a proof_type parameter in a proof object is set to attestation, the object MUST also contain an attestation parameter that includes a JWT as defined in [Section 8.2.1.3](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#attestation-proof-type).
There are two ways to convey key attestations (as defined in [Appendix D](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#keyattestation)) of the cryptographic key material during Credential issuance:
The Wallet uses the jwt proof type in the Credential Request to create a proof of possession of the key and adds the key attestation in the JOSE header.
The Wallet uses the attestation proof type in the Credential Request with the key attestation without a proof of possession of the key itself.
Depending on the Wallet's implementation, the attestation may avoid unnecessary End-User interaction during Credential issuance, as the key itself does not necessarily need to perform signature operations.
Additional proof types MAY be defined and used.

The spec clearly states the content `proofs` for `jwt` and `ldp_vp`, but doesn't mention anything for `attestation`. Additionally, the spec contains only a single example for `proof` when using `attestation`. No examples exists for `proofs` when using `attestation`.

Maybe we should request a clarification from OpenId4VCI.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/IssueCredential.kt
```diff
@@ -467,86 +440,44 @@ private interface Validations : Raise<IssueCredentialError> {
             InvalidProof("You can provide at most '${supportedBatchIssuance.maxProofsSupported}' proofs")
```
dzarras, line 440: @babisRoutis Leaving this here as a reminder. When we add support for Attestation, this check needs to be moved elsewhere. For Attestation we need to ensure that the number of keys being attested is less than `supportedBatchIssuance.maxProofsSupported`. This check cannot be performed here.

babisRoutis (approved):

PR #291, opened 2025-02-19
Advertise supported OpenId4VCI version in metadata and UI pages

### LINKED ISSUE #285: Provide an  indicatation of which draft of the OpenId4VCI spec the app implements
It would be great if we could if application could advertise which draft of the OpenId4VCI is supported.

This could be done in the following way:
- Introduce a custom claim to the credential issuer metadata, and 
- Add to the landing page of the issuer a footer with the same information.

The version of the OpenId4VCI spec could be set by an environment variable

dzarras:
Closes #285

### DIFF
--- a/README.md
+++ b/README.md
@@ -292,6 +292,9 @@ Default value: `10`
 
 ### Metadata configuration
 
+Variable: `ISSUER_METADATA_OPENID4VCI_VERSION`  
+Description: OpenId4VCI version advertised using the `openid4vci_version` property in the Credential Issuer Metadata
+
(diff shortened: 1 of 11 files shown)


babisRoutis:
@dzarras I think that we can also add the version to the credential issuer metadata

dzarras:
> @dzarras I think that we can also add the version to the credential issuer metadata

@babisRoutis It's already there https://github.com/eu-digital-identity-wallet/eudi-srv-pid-issuer/pull/291/files#diff-fec80faa7e6573561cc25dae19efb9930230aca322c57ba4fae2d5eda6e6003cR111

babisRoutis (approved):

PR #307, opened 2025-03-07
Add missing PID attribute in metadata.

### LINKED ISSUE #306: Add missing PID attribute in metadata
The [PID Rulebook v1.5](https://eu-digital-identity-wallet.github.io/eudi-doc-architecture-and-reference-framework/1.5.0/annexes/annex-3/annex-3.01-pid-rulebook/) defines the following optional attributes that we currently do not advertise in the metadata:

* *trust_anchor*: This attribute indicates at least the URL at which a machine-readable version of the trust anchor to be used for verifying the PID can be found or looked up. Note: This attribute corresponds to the location meant in Annex V point h) or Annex VII point h) of the [European Digital Identity Regulation], which is mandatory for QEAAs. This PID Rulebook add this as an optional attribute for PIDs as well, so PID Providers are able to ensure that PIDs can be validated by Relying Parties in the same manner as QEAAs.

The metadata of the issuer must be updated as follows:
Advertise missing *trust_anchor* attribute in MSO MDoc and SD-JWT VC PID Credential Configurations

dzarras:
Add:
1. `trust_anchor` to MSO MDoc and SD-JWT VC PID metadata

Closes #306

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/pid/IssueSdJwtVcPid.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/pid/IssueSdJwtVcPid.kt
@@ -68,6 +68,12 @@ internal object Attributes {
         display = mapOf(Locale.ENGLISH to "Age in Years"),
     )
 
+    val LocationStatus = AttributeDetails(
+        name = "location_status",
+        mandatory = false,
+        display = mapOf(Locale.ENGLISH to "Location Status"),
+    )
+
     val pidAttributes: List<AttributeDetails> = listOf(
         OidcFamilyName,
         OidcGivenName,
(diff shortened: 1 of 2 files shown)


src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/pid/IssueSdJwtVcPid.kt
```diff
@@ -68,6 +68,12 @@ internal object Attributes {
         display = mapOf(Locale.ENGLISH to "Age in Years"),
     )
 
+    val LocationStatus = AttributeDetails(
```
babisRoutis, line 71: Please remove this. 

IMO, the `location_status` will be manifested as `status` in the SD-JWT-VC.

babisRoutis (changes requested):

Issue #322, opened 2025-04-02
Keycloak should advertise OAUTH2 authorization server metadata


babisRoutis:
Strictly speaking OpenId4VCI specification requires that the authorization server protecting a credential issuer should be resolved via well-known location.

In our environment, we use keycloak which is primary an OIDC IdP server. Thus it exposes OIDC metadata rather than OAUTH2 metadata.

To better align with the spec, Keycloak should expose also OAUTH2 metadata.

Hopefully, it seems that there is an easy way to do this

```sh
docker run \
  -p 8080:8080 \
  -e KEYCLOAK_ADMIN=admin \
  -e KEYCLOAK_ADMIN_PASSWORD=admin \
  -e KC_FEATURES=oid4vc-vci \
  quay.io/keycloak/keycloak:26.1.3 \
  start-dev
```

Using the (experimental) feature `oid4vc-vci` (which actually allows Keycloak to act also as a Credential Issuer), adds the OAUTH2 metadata endpoint.

TODO

- [x] Update docker compose
- [x] Expose OAUTH2 authorization metadata to the compliant well-known location (Keycloak just appends well-known locations to the end of the realm `issuer` attribute), via a proxy rule 
- [x] Update PROD and DEV environments

dzarras:
Hi @babisRoutis,

Currently in both DEV and PROD environments we use Keycloak 26.0.5-0. The only enabled feature is dpop.
Both instances advertise OAuth 2.0 Authorization Server Metadata at the following locations:

- https://dev.auth.eudiw.dev/realms/pid-issuer-realm/.well-known/oauth-authorization-server
- https://auth.eudiw.dev/realms/pid-issuer-realm/.well-known/oauth-authorization-server

I'll check whether we can introduce new URLs that are by spec.

I'll start working on updating the local environment as well.

dzarras:
DEV environment has been updated.

Keycloak is now accessible at https://dev.authenticate.eudiw.dev/
OAuth2 Authorization Server Metadata is available at https://dev.authenticate.eudiw.dev/.well-known/oauth-authorization-server/realms/pid-issuer-realm
OpenID Connect Configuration is available at https://[dev.authenticate.eudiw.dev/realms/pid-issuer-realm/.well-known/openid-configuration](https://dev.authenticate.eudiw.dev/realms/pid-issuer-realm/.well-known/openid-configuration)

dzarras:
PROD environment has been updated.

[...]

PR #323, opened 2025-04-03
Expose Authorization Server Metadata via spec aligned URLs

dzarras:
- Upgrades Keycloak to 26.1.4-0
- Exposes OAuth 2.0 Authorization Server Metadata via /.well-known/oauth-authorization-server/idp/realms/pid-issuer-realm
- Exposes OpenID Connect Configuration via /.well-known/openid-configuration/idp/realms/pid-issuer-realm

Tackles #322


babisRoutis (commented):
I think we use the client API of KC to get user details
Shouldn't we update this, as well?

dzarras:
> I think we use the client API of KC to get user details Shouldn't we update this, as well?

@babisRoutis Keycloak admin client doesn't seem to follow the release cadence of Keycloak. I updated the library to its latest version non the less.

babisRoutis (approved):

Issue #348, opened 2025-05-21
Introduce a new API to retrieve Type Metadata


dzarras:
A new API must be introduced in `MetaDataApi` to fetch the Type Metadata of an SD-JWT VC.

This will enable pid-issuer to act as a [trusted Registry](https://www.ietf.org/archive/id/draft-ietf-oauth-sd-jwt-vc-08.html#name-from-a-registry).

The new API must accept as a query parameter the `vct` of the requested Type Metadata.

A new use-case must be implemented that accepts as an input the provided `vct`. The output will be:
- either the request Type Metadata as an `SdJwtVcTypeMetadata` instance (provided by the SD-JWT library)
- or an error indicating the `vct` is not valid per the `Vct` type (provided by the SD-JWT library)
- or an error indicating the Type Metadata of the provided `vct` could not be found
The Type Metadata can be stored in a well known location in the classpath as json files, and the use-case can be configured with a mapping between known `vct` values and the corresponding json file that contains the Type Metadata.

For starters only the `vct` `urn:eudi:pid:1` will be known and served.

The API must be publicly accessible.

PR #349, opened 2025-05-21
Introduce a new API to retrieve Type Metadata

atsant:
Relates #348


src/main/kotlin/eu/europa/ec/eudi/pidissuer/PidIssuerApplication.kt
```diff
@@ -545,12 +547,23 @@ fun beans(clock: Clock) = beans {
     bean {
         CreateCredentialsOffer(ref(), credentialsOfferUri)
     }
+    bean {
+        val knownVct = runCatching {
```
dzarras, line 551: @atsant Some comments:

1. You don't have a single known vct. You have multiple.
2. It's not vctEudi, it the PID vct
3. It's not classPathEudi, it's the PID Type Metadata
4. In case the classpath resource doesn't exist, let it fail, do not let the application start when you have detected configuration errors

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/input/web/MetaDataApi.kt
```diff
@@ -62,10 +70,31 @@ class MetaDataApi(
             .json()
             .bodyValueAndAwait(credentialIssuerMetaData.jwtVcIssuerJwks.toString(true))
 
+    private suspend fun handleGetSdJwtVcTypeMetadata(request: ServerRequest): ServerResponse {
+        val queryParam = request.queryParam("vct")
```
dzarras, line 74: @atsant 

1. try to improve the variable names you use. this is not any query parameter. this is the vct.
2. try to avoid return expressions on assignments. try the following instead:
```kotlin
if (null == vct) {
  return ServerResponse.badRequest()...
}

the rest as is
```

3. No need to set body values for the bad requests.
4. Maybe UnrecognisedVct  better maps to 404/not found

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/input/web/MetaDataApi.kt
```diff
@@ -62,10 +70,31 @@ class MetaDataApi(
...
+                }
+            },
+        )
+    }
+
     companion object {
         const val WELL_KNOWN_OPENID_CREDENTIAL_ISSUER = "/.well-known/openid-credential-issuer"
         const val WELL_KNOWN_JWT_VC_ISSUER = "/.well-known/jwt-vc-issuer"
         const val PUBLIC_KEYS = "/public_keys.jwks"
+        const val SD_JWT_VC_METADATA = "/type-metadata/sd-jwt-vc"
```
dzarras, line 97: @atsant Type Metadata is an SD-JWT VC concept. No need to so explicit. Rename the variable to `TYPE_METADATA` and its value to `/type-metadata`

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/GetTypeMetadata.kt
```diff
@@ -0,0 +1,69 @@
...
+import eu.europa.ec.eudi.sdjwt.vc.SdJwtVcTypeMetadata
+import eu.europa.ec.eudi.sdjwt.vc.Vct
+import kotlinx.serialization.Serializable
+import kotlinx.serialization.json.Json
+import org.springframework.core.io.ClassPathResource
+import java.nio.charset.StandardCharsets
+
+// Sealed interface may be not needed here and subject to removal
+sealed interface MetadataSuccessResponse {
+    @Serializable
```
dzarras, line 31: @atsant No need to introduce this type. Additionally we don't want to return a `Metadata` instance serialized, we want to return an `SdJwtVcTypeMetadata`, without the container.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/GetTypeMetadata.kt
```diff
@@ -0,0 +1,69 @@
...
+    data class Metadata(
+        val metadata: SdJwtVcTypeMetadata,
+    ) : MetadataSuccessResponse
+}
+
+@Serializable
+data class GetTypeMetadataError(val error: String) {
+    companion object {
+        val InvalidVct = GetTypeMetadataError("Provided string could not be converted to VCT")
```
dzarras, line 40: @atsant Simplify the error as `vct is not valid`

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/GetTypeMetadata.kt
```diff
@@ -0,0 +1,69 @@
...
+        val UnrecognisedVct = GetTypeMetadataError("Domain does not recognise VCT")
```
dzarras, line 41: @atsant Rename the error to `UnknownVct` and update the error to `Unknown vct`

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/GetTypeMetadata.kt
```diff
@@ -0,0 +1,69 @@
...
+        val MetadataIncorrectFormat = GetTypeMetadataError("Metadata incorrect format")
```
dzarras, line 42: @atsant Remove this error. We control the contents of the json files that contain the TypeMetadata. If something is incorrect, the application should fail. This is not something to report to the end user.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/GetTypeMetadata.kt
```diff
@@ -0,0 +1,69 @@
...
+
+        MetadataSuccessResponse.Metadata(convertedMetadata)
+    }
+
+    fun Raise<GetTypeMetadataError>.convertToVct(vct: String): Vct =
+        catch({ Vct(vct) }) {
+            raise(GetTypeMetadataError.InvalidVct)
+        }
+
+    fun Raise<GetTypeMetadataError>.convertMetadataToSdJwtVcTypeMetadata(knownClassPath: ClassPathResource): SdJwtVcTypeMetadata =
```
dzarras, line 62: @atsant No need for this specific error. let it fail. if the json is not correct, it's our fault. not something to report to the user.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/GetTypeMetadata.kt
```diff
@@ -0,0 +1,69 @@
...
+) {
+    operator fun invoke(vct: String): Either<GetTypeMetadataError, MetadataSuccessResponse> = either {
+        val vct = convertToVct(vct)
+        val knownClassPath = knownVct.getOrElse(vct) { raise(GetTypeMetadataError.UnrecognisedVct) }
+        val convertedMetadata: SdJwtVcTypeMetadata = convertMetadataToSdJwtVcTypeMetadata(knownClassPath)
+
+        MetadataSuccessResponse.Metadata(convertedMetadata)
+    }
+
+    fun Raise<GetTypeMetadataError>.convertToVct(vct: String): Vct =
```
dzarras, line 57: @atsant try to replace this method with a simple `runCatching { }.getOrElse { raise(...) }`

dzarras:
@atsant We'll also have to update the OpenApi spec and introduce a new group `metadata api` and describe the get request to get the Type Metadata. 
No need to create a full description of the Type Metadata return type, just an object with unknown properties. Check the definition of `PresentationDefinition` for instance.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/PidIssuerApplication.kt
```diff
@@ -546,11 +547,17 @@ fun beans(clock: Clock) = beans {
         CreateCredentialsOffer(ref(), credentialsOfferUri)
     }
 
+    bean {
+        val loadPid = env.getProperty<String>("issuer.typemetadata.available")
+        val knownVct = Json.decodeFromString<Map<Vct, String>>(loadPid!!)
+        GetTypeMetadata(knownVct)
+    }
```
atsant, line 554: @dzarras 
This impl currently has a bug, if the file fails to be converted it will not throw an error at the startup but when the url is called.

Is there anyway we can enforce the whole bean to execute beforehands and crash the application at the startup?

README.md
```diff
@@ -362,12 +362,18 @@ Description: URI where the Wallet can obtain the logo of the Credential Issuer
...
+Each VCT should be configured once.
+
 Variable: `ISSUER_SD_JWT_VC_TYPE_METADATA_XX_VCT` (e.g. `ISSUER_SD_JWT_VC_TYPE_METADATA_0_VCT`)  
-Description: The VCT of the type metadata property  
+Description: The VCT of the type metadata property 
 Example: `urn:eudi:pid:1`  
 
 Variable: `ISSUER_SD_JWT_VC_TYPE_METADATA_XX_RESOURCE` (e.g. `ISSUER_SD_JWT_VC_TYPE_METADATA_0_RESOURCE`)  
-Description: The resource path of the type metadata ARF v1.8 JSON
+Description: Resource of the type metadata, using spring boot class path notation
```
dzarras, line 376: @atsant `using spring framework resource notation`

dzarras (dismissed):

atsant:
@vafeini 
If you can, have a look please.

vafeini:
Looks ok. Personal preference would be to have the vct passed as a path parameter instead of query parameter. From a rest api prespective the specific vct's metadata document it is a sub-resourse of the 'type-metadata' resource.

dzarras:
@atsant Would you like to implement the change that @vafeini  suggested?

dzarras (dismissed):

dzarras:
@atsant Please update the example in README. The Url has been updated and no longer uses path parameters.

PR #355, opened 2025-06-06
Add proxy settings for WebClient

### LINKED ISSUE #354: Add support for configuring an HTTP Proxy
pid-issuer accesses external services during the issuance process.
There are cases where an HTTP Proxy server must be used (e.g. corporate environments).
We must introduce a way for users of pid-issuer to configure an HTTP Proxy server.

Currently pid-issuer uses Spring's WebClient and RestEasy (for Keycloak integration).
We must investigate how these can be configured to use an HTTP Proxy server and add the necessary configuration properties to allow for such configuration.

atsant:
Closes #354

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/pidissuer/PidIssuerApplication.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/pidissuer/PidIssuerApplication.kt
@@ -27,6 +27,7 @@ import com.nimbusds.oauth2.sdk.dpop.verifiers.DefaultDPoPSingleUseChecker
 import com.nimbusds.oauth2.sdk.id.Issuer
 import com.nimbusds.oauth2.sdk.util.X509CertificateUtils
 import com.nimbusds.openid.connect.sdk.op.OIDCProviderMetadata
+import eu.europa.ec.eudi.pidissuer.WebClients.httpClient
 import eu.europa.ec.eudi.pidissuer.adapter.input.web.IssuerApi
 import eu.europa.ec.eudi.pidissuer.adapter.input.web.IssuerUi
 import eu.europa.ec.eudi.pidissuer.adapter.input.web.MetaDataApi
@@ -132,17 +133,12 @@ internal object WebClients {
      * A [WebClient] with [Json] serialization enabled.
      */
     fun default(proxy: HttpProxy?): WebClient {
-        val json = Json { ignoreUnknownKeys = true }
         val httpClient = httpClient(proxy)
         val connector = ReactorClientHttpConnector(httpClient)
         return WebClient
             .builder()
             .clientConnector(connector)
-            .codecs {
-                it.defaultCodecs().kotlinSerializationJsonDecoder(KotlinSerializationJsonDecoder(json))
-                it.defaultCodecs().kotlinSerializationJsonEncoder(KotlinSerializationJsonEncoder(json))
-                it.defaultCodecs().enableLoggingRequestDetails(true)
-            }
+            .configureCodec()
             .build()
     }
 
@@ -158,8 +154,9 @@ internal object WebClients {
             .build()
         val insecureHttpClient = httpClient.secure { it.sslContext(sslContext) }
         val connector = ReactorClientHttpConnector(insecureHttpClient)
-        return default(proxy).mutate()
+        return WebClient.builder()
             .clientConnector(connector)
+            .configureCodec()
             .build()
     }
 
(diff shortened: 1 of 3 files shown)


src/main/kotlin/eu/europa/ec/eudi/pidissuer/PidIssuerApplication.kt
```diff
@@ -147,39 +152,90 @@ internal object WebClients {
...
+    }
+}
+
+data class HttpProxy(
+    val url: Url,
+    val username: String? = null,
+    val password: String? = null,
+) {
+    init {
+        require(password == null || username != null) {
```
dzarras, line 194: @atsant Something we must also verify is that the schema of the url is http, and that the url the user provided has neither path, nor query parameters.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/PidIssuerApplication.kt
```diff
@@ -147,39 +152,90 @@ internal object WebClients {
...
             .loadTrustMaterial(TrustAllStrategy())
             .build()
-        ResteasyClientClassicProvider().newRestEasyClient(null, sslContext, true)
+        val client: Client = ResteasyClientClassicProvider().newRestEasyClient(
+            null,
+            sslContext,
+            true,
+        )
+        return proxy?.let {
+            client.property("org.jboss.resteasy.jaxrs.client.proxy.host", proxy.url)
```
dzarras, line 236: @atsant Please use the constants you defined.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/PidIssuerApplication.kt
```diff
@@ -147,39 +152,90 @@ internal object WebClients {
...
      */
-    val Default: Client by lazy {
-        ResteasyClientClassicProvider().newRestEasyClient(null, null, false)
+    fun default(proxy: HttpProxy?): Client {
+        val client: Client = ResteasyClientClassicProvider().newRestEasyClient(
+            null,
+            null,
+            false,
+        )
+        return proxy?.let {
```
dzarras, line 216: @atsant 
1. See below concerning the syntax.
2. What about the credentials?

dzarras:
@atsant Please rebase against main whenever possible, and let's finalize the required changes.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/PidIssuerApplication.kt
```diff
@@ -145,16 +149,58 @@ internal object WebClients {
...
         log.warn("Using insecure WebClient trusting all certificates")
         val sslContext = SslContextBuilder.forClient()
             .trustManager(InsecureTrustManagerFactory.INSTANCE)
             .build()
-        val httpClient = HttpClient.create().secure { it.sslContext(sslContext) }
-        Default.mutate()
-            .clientConnector(ReactorClientHttpConnector(httpClient))
+        val insecureHttpClient = httpClient.secure { it.sslContext(sslContext) }
+        val connector = ReactorClientHttpConnector(insecureHttpClient)
+        return default(proxy).mutate()
```
dzarras, line 161: It doesn't really make sense to call the `default()` factory.

We are instantiating a `WebClient` and it's connector, just to create a new connector and override it.

In this case it will be a lot cleaner if we simply copy the codecs configuration.

Alternatively, you could also create create an extension method on `WebClient.Builder` to configure the codecs and avoid duplication.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/PidIssuerApplication.kt
```diff
@@ -130,31 +132,72 @@ internal object WebClients {
...
+        val httpClient = httpClient(proxy)
+
         log.warn("Using insecure WebClient trusting all certificates")
         val sslContext = SslContextBuilder.forClient()
             .trustManager(InsecureTrustManagerFactory.INSTANCE)
             .build()
-        val httpClient = HttpClient.create().secure { it.sslContext(sslContext) }
-        Default.mutate()
-            .clientConnector(ReactorClientHttpConnector(httpClient))
+        val insecureHttpClient = httpClient.secure { it.sslContext(sslContext) }
```
dzarras, line 155: You can define `httpClient` as follows:

```kotlin
val httpClient = httpClient(proxy).secure { it.sslContext(sslContext) }
```

We don't need both a secure and insecure client.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/PidIssuerApplication.kt
```diff
@@ -130,31 +132,72 @@ internal object WebClients {
...
+                    proxy.username?.let {
+                        username(it)
+                        password { proxy.password ?: "" }
+                    }
+                }
+        }
+    }
+}
+
+data class HttpProxy(
```
dzarras, line 182: Please make this a top-level class. Similarly to `KeycloakConfigurationProperties` for instance. 
No need to nest this inside `WebClients`.

atsant: @dzarras Currently it is not nested inside `WebClients`, but can be moved near `KeycloakConfigurationProperties` if needed

PR #383, opened 2025-07-17
Support key attestations in JWT proofs and `attestation` proof type

### LINKED ISSUE #277: Support key attestations in JWT proofs

### LINKED ISSUE #287: Support `attestation` proof
Draft 15 intoruces the concept of "key attestation" as a way for the holder to provide evidence for the authenticity and security properties of a key and its storage component to the Credential Issuer.

Spec forsees that a key attestation jwt can be used as proof of possession of key material to be embedded in issued credentials (see [here](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-15.html#section-8.2.1.3))

It also mentiones that 

>The Credential Issuer SHOULD issue a Credential for each cryptographic public key specified in the attested_keys claim within the key_attestation parameter.

For the above to be supported issuer must:
- [ ] Introduce new proof type `attestation` in credential configuration's `proof_types_supported`
- [ ] Validate the key attestation passed from wallet:
	- Validate that key attestation is required from the credential's policy (configuration includes it in `proof_types_supported`)
 	- Validate signature of key attestation as described in #277 
	- Validate that key attestation is aligned with the specs in credential configuration ([key_attestations_required.key_storage](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-15.html#section-11.2.3-2.11.2.5.2.2.2.1) and [key_attestations_required.user_authentication](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-15.html#section-11.2.3-2.11.2.5.2.2.2.2) )
- [ ] For every key included in key attestation's `attested_keys` claim a separate credential instance should be created for the same dataset and format.

vafeini:
Closes #277
Closes #287


src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/KeyAttestationJWT.kt
```diff
@@ -0,0 +1,143 @@
...
+                "Invalid Key Attestation JWT. Item at index $index in `attested_keys` is not a JSON object."
+            }
+            try {
+                @Suppress("UNCHECKED_CAST")
+                val jwk = JWK.parse(keyObject as Map<String, Any>)
+                require(!jwk.isPrivate) {
+                    "Invalid Key Attestation JWT. Item at index $index in `attested_keys` must be a public key."
+                }
+                jwk
+            } catch (e: ParseException) {
```
dzarras, line 59: There is a potential `ClassCastException` here as well. If the Key Attestation JWT is malformed, `attested_keys` might contain arbitrary values.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/KeyAttestationJWT.kt
```diff
@@ -0,0 +1,143 @@
...
+    companion object {
+        const val KEY_ATTESTATION_JWT_TYPE = "keyattestation+jwt"
+
+        operator fun invoke(value: String): KeyAttestationJWT = KeyAttestationJWT(SignedJWT.parse(value))
+        operator fun invoke(signedJWT: SignedJWT): KeyAttestationJWT = KeyAttestationJWT(signedJWT)
+    }
+}
+
+fun KeyAttestationJWT.ensureValidKeyAttestation(keyAttestationRequirement: KeyAttestation) {
+    require(keyAttestationRequirement is KeyAttestation.Required) {
```
dzarras, line 104: Why not limit `keyAttestationRequirement` to `KeyAttestation.Required` then?

Also the error message makes sense only when this function is called when validating a JWT Proof. This is used though to validate Attestation Proofs as well.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/ValidateProofs.kt
```diff
@@ -64,7 +65,10 @@ internal class ValidateProofs(
                 extractJwkFromCredentialKey(it.first).getOrElse { error ->
                     raise(InvalidProof("Unable to extract JWK from CredentialKey", error))
                 }
-            }.toNonEmptyListOrNull()
+            }.flatten()
+                .distinct()
```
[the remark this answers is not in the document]
vafeini: To be honest I think this will make issuer too strict. The way the spec is now does not prohibit callers to send duplicated keys.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/ValidateJwtProof.kt
```diff
@@ -121,6 +129,12 @@ private fun CredentialKey.ensureCompatibleWith(algorithm: JWSAlgorithm) {
         is CredentialKey.DIDUrl -> jwk.ensureCompatibleWith(algorithm)
         is CredentialKey.Jwk -> value.ensureCompatibleWith(algorithm)
 
+        is CredentialKey.AttestedKeys -> {
+            val signingJWK = keys.signingKeyOf(signedJwt)
+            requireNotNull(signingJWK) { "Key attestation does not contain a key that verifies the jwt proof signature" }
```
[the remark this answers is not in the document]
vafeini: No this is not the puprose of this check. Here we check that the key that signed the JWT proof is compatible with the algorithm of the JWT Proof header. 
The only way we have so far to find the key that signed the JWT Proof is to go over the list of attested_keys and validate signature with each one. 
There is a non-normative example in the spec [here](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-15.html#section-8.2.1.1-11) where the `kid` is used to identify the index of the key in the `attested_keys` array but this is just an example, nothing in the spec itself that we can be based upon.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/KeyAttestationJWT.kt
vafeini: Not sure I understand. There is a require(keyObject is Map<*, *>) for each element of the array and if element is not parsable as JWK an exception will thrown

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/ValidateJwtProof.kt
```diff
@@ -91,19 +90,37 @@ private fun algorithmAndCredentialKey(
...
+private fun CredentialKey.AttestedKeys.Companion.fromKeyAttestation(
+    keyAttestation: String,
+    keyAttestationRequirement: KeyAttestation,
+): CredentialKey.AttestedKeys {
+    require(keyAttestationRequirement is KeyAttestation.Required) {
+        "Proof type JWT does not require key attestation, though one was provided."
+    }
+    val keyAttestationJWT = KeyAttestationJWT(keyAttestation).also {
+        it.ensureValidKeyAttestation(keyAttestationRequirement)
+    }
```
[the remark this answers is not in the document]
vafeini: Yes this is correct, I will add it

gdimtsas (dismissed):

src/main/kotlin/eu/europa/ec/eudi/pidissuer/PidIssuerApplication.kt
```diff
@@ -459,9 +460,18 @@ fun beans(clock: Clock) = beans {
     //
     // Specific Issuers
     //
-    bean { ValidateJwtProof(issuerPublicUrl) }
+    bean(name = "verify-key-attestation-with-exp-claim") {
```
babisRoutis, line 463: We can have a single bean 

And pass the option to validate `exp` or not as an argument to the method  `invoke`

That is, let the caller decide

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/pid/IssueSdJwtVcPid.kt
```diff
@@ -184,9 +184,31 @@ internal val SdJwtVcPidVct: SdJwtVcType = SdJwtVcType(pidDocType(1))
...
+    keyAttestationRequirement: KeyAttestation,
+): Set<ProofType> = buildSet {
+    add(
+        ProofType.Jwt(
+            jwtProofsSupportedSigningAlgorithms,
+            keyAttestationRequirement,
+        ),
+    )
+    // Attestation proof is available only when key attestations for this credential are enabled in configuration
+    if (keyAttestationRequirement is KeyAttestation.Required) {
```
babisRoutis, line 198: Please replace if expression with a when (without else)
Make pattern matching total instead of partial

src/main/resources/application.properties
```diff
@@ -47,6 +47,9 @@ issuer.ehic.encoder.integrity.hashAlgorithm=sha-256
 issuer.ehic.notifications.enabled=true
 issuer.ehic.issuingCountry=GR
 issuer.ehic.jwtProofs.supportedSigningAlgorithms=ES256
+issuer.ehic.key_attestations.required = true
+issuer.ehic.key_attestations.constraints.key_storage = iso_18045_high, iso_18045_moderate, iso_18045_enhanced-basic, iso_18045_basic
```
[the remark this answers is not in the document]
vafeini: It is unclear to me if there is some kind of hierarchy between the values (ex. if issuer demands `iso_18045_enhanced-basic` then its acceptable to have `iso_18045_moderate` or `iso_18045_high` in attestation). For the time being as spec defines that these attributes are arrays with multiple values, I have added all values. We need to revisit this.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/VerifyKeyAttestation.kt
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    private fun KeyAttestationJWT.extractSigningKey(): JWK {
+        val header = jwt.header
+        val kid = header.keyID
+        val x5c = header.x509CertChain
+
+        return when {
+            kid != null && x5c.isNullOrEmpty() -> resolveDidUrl(URI.create(kid)).getOrThrow()
+            kid == null && !x5c.isNullOrEmpty() -> {
```
babisRoutis, line 86: Remove the `kid == null`. We don't care

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/VerifyKeyAttestation.kt
```diff
@@ -0,0 +1,214 @@
...
+                trustAnchors?.let {
```
babisRoutis, line 87: I think an if would be more readable

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/VerifyKeyAttestation.kt
```diff
@@ -0,0 +1,214 @@
...
+                    val chain = x5c.map { X509CertUtils.parse(it.decode()) }
+                    chain.isTrusted(it)
+                }
+                parseDer(x5c).getOrThrow()
```
babisRoutis, line 91: Why are we parsing again?


Please consider the following 

```kotlin
val chain : NonEmptyList<X509Cert> = TODO() 
if (trustAnchors != null) {
    // validate 
}
chain(0)

```

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/VerifyKeyAttestation.kt
```diff
@@ -0,0 +1,214 @@
...
+            verifyCNonce(at)
+            verifySignature(key, algorithm)
+            ensureMeetsKeyAttestationRequirements(keyAttestationRequirement)
+        }
+        keyAttestation.attestedKeys
+    }
+
+    private fun KeyAttestationJWT.extractSigningKey(): JWK {
+        val header = jwt.header
+        val kid = header.keyID
```
babisRoutis, line 81: Please specify types explicitly (declaring them as nullable)

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/VerifyKeyAttestation.kt
```diff
@@ -0,0 +1,214 @@
...
+) {
+    suspend operator fun invoke(
+        keyAttestation: KeyAttestationJWT,
+        signingAlgorithmsSupported: NonEmptySet<JWSAlgorithm>,
+        keyAttestationRequirement: KeyAttestation.Required,
+        at: Instant,
+    ): Either<Throwable, NonEmptyList<JWK>> = Either.catch {
+        with(keyAttestation) {
+            val algorithm = extractSupportedAlgorithm(signingAlgorithmsSupported)
+            val key = extractSigningKey().also {
```
babisRoutis, line 63: Use `apply` instead of also.

Also consider the following

```kotlin
fun JWK.ensureCompatibleWith() : JWK = apply {

}
```

With this style you can just "wire" with dot 

```kotlin

extractSigningKey().ensureCompatibleWith()
```

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/pid/IssueMsoMdocPid.kt
```diff
@@ -274,7 +273,31 @@ val PidMsoMdocV1CredentialConfigurationId: CredentialConfigurationId = Credentia
 
 val PidMsoMdocV1DocType: MsoDocType = pidDocType(1)
 
-fun pidMsoMdocV1(jwtProofsSupportedSigningAlgorithms: NonEmptySet<JWSAlgorithm>): MsoMdocCredentialConfiguration =
+internal fun pidMsoMdocProofTypesSupported(
```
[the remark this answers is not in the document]
vafeini: Having them defined separately per attestation issuer gives the flexibility to configure different sets of proofs per attestation. If we don't care we can move that to ProofType companion object

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/VerifyKeyAttestation.kt
```diff
@@ -0,0 +1,214 @@
...
+import java.time.Instant
+import kotlin.time.Duration
+import kotlin.time.Duration.Companion.seconds
+import kotlin.time.DurationUnit
+
+internal val SkipRevocation: PKIXParameters.() -> Unit = { isRevocationEnabled = false }
+
+internal class VerifyKeyAttestation(
+    private val trustAnchors: KeyStore? = null,
+    private val verifyAttestedKey: (JWK) -> JWK?,
```
babisRoutis, line 50: The abstraction is problematic

- it is lossy. Totally discards the error channel.
- it is not suspend function
- Most importantly, it doesn't take into account the attestation requirements as defined by the issuer


Please consider introducing a function interface 

```kotlin
fun interface VerifyAttestedKey {
    suspend fun verify(key: JWK, requirement: KeyAttestation.Required): Either<SomeError, JWK>
    suspend fun verify(keys: NomEmptyList<JWK>, requirement: KeyAttestation.Required): Either<SomeError, JWK> = TODO ("Implement this using arrow, accumulating errors")
}


```

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/VerifyKeyAttestation.kt
```diff
@@ -0,0 +1,214 @@
...
+        .takeIf(JWSAlgorithm.Family.SIGNATURE::contains)
+        ?.takeIf(signingAlgorithmsSupported::contains)
+        ?: error("signing algorithm of key attestation '${jwt.header.algorithm.name}' is not supported")
+
+private fun JWK.ensureCompatibleWith(algorithm: JWSAlgorithm) {
+    val supportedAlgorithms =
+        when (this) {
+            is RSAKey -> RSASSASigner.SUPPORTED_ALGORITHMS
+            is ECKey -> ECDSASigner.SUPPORTED_ALGORITHMS
+            is OctetKeyPair -> Ed25519Signer.SUPPORTED_ALGORITHMS
```
babisRoutis, line 169: Why support `OctetKeyPair` ?
This needs special handling.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/VerifyKeyAttestation.kt
```diff
@@ -0,0 +1,214 @@
...
+private fun KeyAttestationJWT.extractSupportedAlgorithm(signingAlgorithmsSupported: NonEmptySet<JWSAlgorithm>): JWSAlgorithm =
+    jwt.header.algorithm
+        .takeIf(JWSAlgorithm.Family.SIGNATURE::contains)
+        ?.takeIf(signingAlgorithmsSupported::contains)
+        ?: error("signing algorithm of key attestation '${jwt.header.algorithm.name}' is not supported")
+
+private fun JWK.ensureCompatibleWith(algorithm: JWSAlgorithm) {
+    val supportedAlgorithms =
+        when (this) {
+            is RSAKey -> RSASSASigner.SUPPORTED_ALGORITHMS
```
babisRoutis, line 167: Why RSA?

Is there a requirement for this?

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/VerifyKeyAttestation.kt
babisRoutis: Check also my comment about `cNonce`

Please do not remove it. Just  define the interface

src/main/resources/application.properties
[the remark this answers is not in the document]
babisRoutis: @vafeini They are optional. Please default to empty

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/pid/IssueMsoMdocPid.kt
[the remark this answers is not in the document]
babisRoutis: I am not sure that I follow your point 

A specic credential configurations calls this single method with whatever policy is applicable to the configuration.
What is this has to do with implemented the same method several times?

babisRoutis (changes requested):
Dear @vafeini please consider my comments.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/VerifyKeyAttestation.kt
vafeini: Removed support to RSA algorithms and keys from JWT and attestation proofs

Issue #391, opened 2025-08-08
Signed Issuer Metadata


vafeini:
Regarding the GET request to fetch issuer metadata draft 16 mentions the following

> The Wallet is RECOMMENDED to send an Accept header in the HTTP GET request to indicate the Content Type(s) it supports, and by doing so, signaling whether it supports signed metadata.
>
> The Credential Issuer MUST respond with HTTP Status Code 200 and return the Credential Issuer Metadata containing the parameters defined in [Section 11.2.4](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-16.html#credential-issuer-parameters) as either
> 
> - an unsigned JSON document using the media type application/json, or
> - a signed JSON Web Token (JWT) containing the Credential Issuer Metadata in its payload using the media type application/jwt.
>
> The Credential Issuer MUST support returning metadata in an unsigned form 'application/json' and MAY support returning it in a signed form 'application/jwt'. In all cases the Credential Issuer MUST indicate the media type of the returned Metadata using the HTTP Content-Type header. It is RECOMMENDED for Credential Issuers to respond with a Content-Type matching to the Wallet's requested Accept header when the requested content type is supported.

- [x] Depending on the ACCEPT header of the GET request for issuer metadata return metadata as plain json or signed JWT. 
- [x] Generate signed metadata JWT as specified [here](https://openid.github.io/OpenID4VCI/openid-4-verifiable-credential-issuance-wg-draft.html#name-signed-metadata) (this is a link to draft 17 the structure is better defined there)
- [x] Include the respective Content-Type header in the response

Issue #390, opened 2025-08-08
Issuer metadata updates


vafeini:
The following changes introduces to issuer metadata in draft 16:

- [ ] Introduce `credential_request_encryption`
- [ ] Add `zip_values_supported` in `credential_response_encryption`
- [ ] Remove `signed_metadata` property
- [ ] Updates `credential_configurations_supported`:
        -  Values of `credential_signing_alg_values_supported` are type specific (COSE algorithms used for msoMdoc formated credentials)
        - `cryptographic_binding_methods_supported` MUST be present when key binding is required for a credential, and omitted otherwise. If absent, key binding is not required for this credential
        - Property `proof_types_supported` must be present if `cryptographic_binding_methods_supported` is present 
        - Add `credential_metadata` property grouping `display` and `claims` properties
- [ ] Rename `ldp_vp` to `di_vp`

PR #403, opened 2025-08-27
Align issuer metadata with draft 16

gdimtsas:
Tackles #390
Tackles #391


src/main/kotlin/eu/europa/ec/eudi/pidissuer/PidIssuerApplication.kt
```diff
@@ -916,6 +918,42 @@ private fun BeanDefinitionDsl.keyAttestationRequirement(attestationPropertyPrefi
...
 }
 
+private fun Environment.credentialRequestEncryption(): CredentialRequestEncryption {
+    val isSupported = getProperty<Boolean>("issuer.credentialRequestEncryption.supported") ?: false
+    return if (!isSupported) {
+        CredentialRequestEncryption.NotSupported
+    } else {
+        val key = when (getProperty<KeyOption>("issuer.credentialRequestEncryption.jwks")) {
+            null, KeyOption.GenerateRandom -> {
+                ECKeyGenerator(Curve.P_256)
```
dzarras, line 928: @gdimtsas 

1. Let's use a different key id. `issuer-kid-1` is also used for the signing key as well. Maybe we should be using a UUID instead.
2. Per https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-16.html#name-encrypted-messages the JWK must contain the `alg` claim, indicating which JWE Algorithm the JWK must be used with. We are currently missing this. - Alternatively this could be added in `CredentialRequestEncryptionSupportedParameters`, doing so would also enforce these when loading a JWK from a Keystore.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/CredentialIssuerMetaData.kt
```diff
@@ -16,11 +16,39 @@
...
+ * for encryption of the Credential Request
+ * @param methodsSupported a list of the JWE RFC7516 encryption algorithms
+ * (enc values) RFC7518 supported by the Credential
+ * and/or Batch Credential Endpoint to decode the Credential or
+ * Batch Credential Response in a JWT RFC7519
+ * @param zipAlgorithmsSupported a list of the JWE RFC7516 compression
+ * algorithms (zip values) RFC7518 supported by the Credential Endpoint
+ * to uncompress the Credential Request after decryption
+ */
+data class CredentialRequestEncryptionSupportedParameters(
```
dzarras, line 40: @gdimtsas I'd suggest to rename `methodsSupported` to `encryptionMethodsSupported` and `zipAlgorithmsSupported` to `compressionAlgorithmsSupported`, to better align with the rest of the codebase.

dzarras: On second thought, we can skip this. I see we use both `methodsSupported` and `encryptionMethod` as terms in the codebase. Up to you.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/CredentialConfiguration.kt
```diff
@@ -130,3 +130,11 @@ sealed interface CredentialConfiguration {
     val credentialSigningAlgorithmsSupported: Set<JWSAlgorithm>
     val proofTypesSupported: ProofTypesSupported
 }
+
+internal fun CredentialConfiguration.validateCryptographicBindingsAndProofTypes() {
+    val hasCryptoBinding = cryptographicBindingMethodsSupported.isNotEmpty()
+    val hasProofTypes = proofTypesSupported != ProofTypesSupported.Empty
+    require(hasCryptoBinding == hasProofTypes) {
+        "proofTypesSupported be present if cryptographicBindingMethodsSupported are provided, and omitted otherwise"
```
dzarras, line 138: @gdimtsas small type. `"proofTypesSupported must be present...`

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/CredentialConfiguration.kt
```diff
@@ -130,3 +130,11 @@ sealed interface CredentialConfiguration {
     val credentialSigningAlgorithmsSupported: Set<JWSAlgorithm>
```
dzarras, line 130: @gdimtsas Concerning `credentialSigningAlgorithmsSupported`: I think we should not force the usage of `JWSAlgorithm` in the interface. MSO MDoc uses COSE values, which are not compatible with JOSE.

I was thinking about something like this maybe:

```kotlin
@JvmInline
value class CoseAlgorithm(val value: Int)

sealed interface CredentialSigningAlgorithm {

    @JvmInline
    value class Jose(val value: JWSAlgorithm) : CredentialSigningAlgorithm

    @JvmInline
    value class Cose(val value: CoseAlgorithm) : CredentialSigningAlgorithm
}
```

Have `CredentialConfiguration` define `credentialSigningAlgorithmsSupported` as a `Set<CredentialSigningAlgorithm>` and then further refine the type in each subclass.

So `MsoMdocCredentialConfiguration` would use `CredentialSigningAlgorithm.Cose` and `SdJwtVcCredentialConfiguration` would use `CredentialSigningAlgorithm.Jose`

cc: @babisRoutis

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/GetCredentialIssuerMetaData.kt
```diff
@@ -68,12 +62,26 @@ data class CredentialIssuerMetaDataTO(
     val openid4VciVersion: String? = null,
 ) {
 
+    @Serializable
+    data class CredentialRequestEncryptionTO(
+        @Required @SerialName("jwks")
+        val jwks: JsonArray,
```
dzarras, line 68: @gdimtsas This is not correct. `jwks` is a serialized JWK Set, not a list of JWKs. Check: https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-16.html#section-11.2.4-2.7.2.1

dzarras: Better use `JsonObject` here. Then later you can convert using: 
```kotlin
Json.decodeFromString(JSONObjectUtils.toJSONString(optional.parameters.jwks.toJSONObject()))
```

dzarras: Check here as well: https://github.com/eu-digital-identity-wallet/eudi-lib-jvm-openid4vci-kt/pull/448/files#diff-fe88f490e9286f33b71b7799effcfc0aabace43b4b618c654dcefd0d44dcff28R467

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/CredentialIssuerMetaData.kt
gdimtsas: I'm thinking that we should keep it as is for consistency

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/input/web/MetaDataApi.kt
```diff
@@ -38,7 +40,10 @@ class MetaDataApi(
 
     val route = coRouter {
         GET(WELL_KNOWN_OPENID_CREDENTIAL_ISSUER, accept(MediaType.APPLICATION_JSON)) { _ ->
-            handleGetClientIssuerMetaData()
```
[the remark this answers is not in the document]
gdimtsas: It's already pushed, specifically this part (lines 63-66):

```kotlin
private suspend fun handleGetSignedCredentialIssuerMetaData(): ServerResponse =
        getCredentialIssuerMetaData.signed()?.let { metaData ->
            ServerResponse.ok().contentType(MEDIA_TYPE_APPLICATION_JWT).bodyValueAndAwait(metaData)
        } ?: handleGetUnsignedCredentialIssuerMetaData()
```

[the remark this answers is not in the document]
dzarras: Sorry my bad. Hadn't noticed it. Looks good.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/PidIssuerApplication.kt
```diff
@@ -1036,12 +1078,23 @@ private fun loadJwkFromKeystore(environment: Environment, prefix: String): JWK {
     val keystorePassword = environment.getProperty(property("keystore.password"))?.takeIf { it.isNotBlank() }
     val keyAlias = environment.getRequiredProperty(property("alias"))
     val keyPassword = environment.getProperty(property("password"))?.takeIf { it.isNotBlank() }
+    val keyAlgorithm = environment.getProperty(property("algorithm"))?.takeIf { it.isNotBlank() }
```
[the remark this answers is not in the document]
gdimtsas: I updated it to Algorithm, since the key buildes below, where the key algorithm is used, also expect an Algorithm as a param.

src/main/resources/application-keystore.properties
```diff
@@ -19,4 +19,4 @@ issuer.credentialRequestEncryption.jwks.keystore.type=JKS
 issuer.credentialRequestEncryption.jwks.keystore.password=
 issuer.credentialRequestEncryption.jwks.alias=request-encryption
 issuer.credentialRequestEncryption.jwks.password=request-encryption
-issuer.credentialRequestEncryption.jwks.algorithm=ES256
\ No newline at end of file
+issuer.credentialRequestEncryption.jwks.algorithm=ECDH-ES+A128KW
```
dzarras, line 22: @gdimtsas Let's change this to `ECDH-ES` instead.

dzarras (approved):

PR #409, opened 2025-09-01
Issuance request/response updates

### LINKED ISSUE #395: Issuance and deferred issuance request payload updates
The following updates have been introduced regarding issuance request payload:

- [ ] Removed `proof` attribute from issuance request. Its equivalent is a `proofs` array with one proof included
- [ ] Updated `credential_response_encryption` as specified [here](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-16.html#section-8.2-2.4.1)

The following updates have been introduced regarding deferred issuance request payload:

- [ ] Include `credential_response_encryption` in request payload to specify encryption specifications. Encryption params are defined anew when deferred issuance request is placed and don't have to match the encryption params included in the initial issuance request.

gdimtsas:
Closes #395


src/main/kotlin/eu/europa/ec/eudi/pidissuer/PidIssuerApplication.kt
```diff
@@ -539,8 +539,10 @@ fun beans(clock: Clock) = beans {
 
                     val deferred = env.getProperty<Boolean>("issuer.pid.sd_jwt_vc.deferred") ?: false
                     add(
-                        if (deferred) issueSdJwtVcPid.asDeferred(ref(), ref())
-                        else issueSdJwtVcPid,
+                        if (deferred) {
+                            val interval = env.getProperty<Long>("issuer.pid.sd_jwt_vc.deferred.interval", 1000L)
```
dzarras, line 543: @gdimtsas Please use `Duration` instead. Check for instance how `issuer.pid.sd_jwt_vc.duration` is parsed.

PR #410, opened 2025-09-03
Improve reliability and maintainability

atsant:


src/main/kotlin/eu/europa/ec/eudi/pidissuer/PidIssuerApplication.kt
```diff
@@ -249,7 +249,8 @@ fun beans(clock: Clock) = beans {
     //
     // CNonce encryption key
     //
-    bean(name = "cnonce-encryption-key") {
+    val CNONCE_ENCRYPTION_KEY = "cnonce-encryption-key"
+    bean(name = CNONCE_ENCRYPTION_KEY) {
```
dzarras, line 253: @atsant The cnonce encryption doesn't have to be defined as a bean. We can simply assign the result of `when` to a `val` and instead reference the `val` in the `GenerateCNonceAndEncryptWithNimbus` and `DecryptCNonceWithNimbusAndVerify` bean definitions.

PR #415, opened 2025-09-05
Implement encrypted credential requests

### LINKED ISSUE #396: Encrypted issuance and deferred issuance requests
The following apply to issuance and deferred issuance requests:

- [ ] Issuance requests must be encrypted by the spec provided from issuer [here](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-16.html#section-11.2.4-2.7.1) 
- [ ] Encryption must be done according to section [9.4](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0-16.html#section-9.4) of the spec.
- [ ] Investigate how compression can be applied to ecrypted requests before encrypting respecting `zip_values_supported` specified from issuer.

For issuance and deferred issuance requests, spec mentions the following

> Credential Request encryption MUST but used if the credential_response_encryption parameter is included, to prevent it being substituted by an attacker.

For this to happen when issuer includes in its metadata `credential_response_encryption` it must also provide `credential_request_encryption`. In other words if issuer supports encrypted responses it must advertice in metadata its capabilities for encrypted requests. 

- [ ] Issuer metadata must include `credential_request_encryption` when `credential_response_encryption` is included.
- [ ] Expect that issuance requests (deferred or not) are encrypted if `credential_response_encryption` is included in the request.
- [ ] Expect that issuance requests (deferred or not) are encrypted when issuer metadata have `credential_request_encryption.encryption_required` set to true

gdimtsas:
Closes #396

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/IssueCredential.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/IssueCredential.kt
@@ -25,8 +25,8 @@ import com.nimbusds.jose.EncryptionMethod
 import com.nimbusds.jose.jwk.source.ImmutableJWKSet
 import com.nimbusds.jose.proc.JWEDecryptionKeySelector
 import com.nimbusds.jose.proc.SecurityContext
+import com.nimbusds.jose.util.JSONObjectUtils
 import com.nimbusds.jwt.EncryptedJWT
-import com.nimbusds.jwt.JWTClaimsSet
 import com.nimbusds.jwt.proc.DefaultJWTProcessor
 import eu.europa.ec.eudi.pidissuer.adapter.out.util.getOrThrow
 import eu.europa.ec.eudi.pidissuer.domain.*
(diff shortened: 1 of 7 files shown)


dzarras:
@gdimtsas Please also rebase against the updated release branch.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/IssueCredential.kt
```diff
@@ -272,6 +326,37 @@ class IssueCredential(
...
+                    requestEncryption.parameters
+                }
+                is CredentialRequestEncryption.NotSupported -> raise(RequestEncryptionNotSupported)
+            }
+
+        ensure(methodsSupported.any { it == jwt.header.encryptionMethod}) {
+            UnsupportedEncryptionMethod(jwt.header.encryptionMethod, methodsSupported)
+        }
+        jwt.header.compressionAlgorithm?.let { compressionAlgorithm ->
+            ensure(compressionMethodsSupported?.any { it.name == compressionAlgorithm.name } == true) {
```
dzarras, line 345: @gdimtsas Let's split this into two distinct checks:

```kotlin
ensureNotNull(compressionMethodsSupported) {
     UnsupportedCompressionMethod(compressionAlgorithm, null)
}
ensure(compressionMethodsSupported.any { it == compressionAlgorithm }) {
    UnsupportedCompressionMethod(compressionAlgorithm, compressionMethodsSupported)
}
```

PR #451, opened 2025-11-10
Update handling of proofs and nonce

### LINKED ISSUE #449: Make `nonce` of `key_attestation` optional, when using Proof `jwt`
Given that our issuer advertises `nonce_endpoint`, `nonce`/`proof` handling/validation must be updated as follows:

* `jwt` proofs must always contain a `nonce` that is a valid issuer-provided `c_nonce`
    * if a `jwt` proof contains a `key_attestation`, the `key_attestation` might **optionally** contain a `nonce` that must be a valid issuer-provided `c_nonce`. ~the `nonce` in `key_attestation` does not necessarily have to match the `nonce` of the `jwt` proof.~
* `attestation` proofs must always contain a `nonce` that is a valid issuer-provided `c_nonce`

Reference: https://github.com/openid/OpenID4VCI/issues/676#issuecomment-3501443791

dzarras:
Updates handling of `proofs` and `nonce`.

1. Removes verification of `nonce` from `VerifyKeyAttestation`. Whether the Signed JWT under verification is a valid Key Attestation from a Wallet Provider, and whether it contains a `nonce` generated by pid-issuer are separate concerns.
2. Both `ValidateJwtProof`, and `ValidateAttestationProof` enforce the presence of a `nonce`. Still the verification of `nonce` is not their concern. `nonce` verification is done once in `ValidateProofs`.
3. When `ValidateJwtProof` verifies a JWT Proof that contains a Key Attestation, if the Key Attestation contains a `nonce`, it now enforces it to be equal to the `nonce` of the JWT Proof.

Closes #449

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/input/web/security/DPoPTokenReactiveAuthenticationManager.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/input/web/security/DPoPTokenReactiveAuthenticationManager.kt
@@ -177,6 +177,6 @@ private suspend fun DPoPNoncePolicy.verifyDPoPNonce(unverified: String?, now: In
             else null
 
         is DPoPNoncePolicy.Enforcing ->
-            if (verifyDPoPNonce(unverified, now)) Nonce(unverified)
+            if (unverified != null && verifyDPoPNonce(unverified, now)) Nonce(unverified)
             else throw OAuth2AuthenticationException(DPoPTokenError.useDPoPNonce("Invalid DPoP proof."))
     }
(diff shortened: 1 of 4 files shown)


src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/ValidateAttestationProof.kt
```diff
@@ -56,6 +56,7 @@ internal class ValidateAttestationProof(
             expectExpirationClaim = false,
             at = at,
         ).getOrThrow()
+        requireNotNull(nonce) { "Key attestation does not contain a c_nonce." }
```
babisRoutis: I have the impression that here we have to use `VerifyNonce`.
Right?

babisRoutis (approved):

PR #456, opened 2025-11-13
Implement configuring a separate signing key per credential configuration

### LINKED ISSUE #455: Add support for configuring Signing Key/Certificate per Credential Configuration
PID Issuer currently uses a single signing key/certificate for all Credential Configurations.

Given that different Credential types need to use different Certificate Profiles, we must enhance PID Issuer to allow configuring a signing key/certificate per Credential Configuration.

dzarras:
This PR introduces the ability to configure a separate signing key per Credential Configuration.

The only exception is EHIC. Currently both Credential Configurations of EHIC are configured using the same configuration properties. For uniformity we also use the same signing key for both.

Closes #455

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/pidissuer/PidIssuerApplication.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/pidissuer/PidIssuerApplication.kt
@@ -290,9 +290,15 @@ fun beans(clock: Clock) = beans {
 
                     when (val loadedJwk = issuerKeystore.loadJwk(env, "issuer.credentialRequestEncryption.jwks")) {
                         is ECKey -> {
+                            require(keyAlgorithm in loadedJwk.supportedJWEAlgorithms) {
+                                "${keyAlgorithm.name} cannot be used with an ECKey"
+                            }
                             ECKey.Builder(loadedJwk).algorithm(keyAlgorithm).build()
                         }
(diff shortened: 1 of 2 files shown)


babisRoutis:
@dzarras 

Have some questions:

1. Do we really need to keep the option to generate signing keys?
2. Do we need separate key-stores ?

src/main/kotlin/eu/europa/ec/eudi/pidissuer/PidIssuerApplication.kt
```diff
@@ -252,7 +330,7 @@ fun beans(clock: Clock) = beans {
 
             KeyOption.LoadFromKeystore -> {
                 log.info("Loading Nonce encryption key from keystore")
-                val nonceEncryptionKey = loadJwkFromKeystore(env, "issuer.nonce.encryption-key")
+                val nonceEncryptionKey = issuerKeystore.loadJwk(env, "issuer.nonce.encryption-key")
```
vafeini: I think we are missing the issuer.nonce.encryption-key from properties file

dzarras: They are documented: https://github.com/eu-digital-identity-wallet/eudi-srv-pid-issuer/blob/main/README.md?plain=1#L502

PR #463, opened 2025-12-01
Add support for Protected Resource Metadata

### LINKED ISSUE #441: Expose Protected Resource Metadata
Consider exposing Protected Resource Metadata according to [RFC9728](https://www.rfc-editor.org/rfc/rfc9728.html)

There is no such a requirement to OpenId4VCI, yet it would provide a convenient way to document the decisions made by our implementation.

Possible attributes to use
-  `resource` : The same value as `credential_issuer` from Credential Issuer metadata from OpenId4VCI 
- `authorization_servers` : Identical to the `authorization_servers` from Credential Issuer metadata from OpenId4VCI 
- `scopes_supported` : An array for the scopes defined by our Credential Issuer (to access Credential endpoint)
- `bearer_methods_supported` : [`header`]  Assuming that we will continue to support bearer authentication. Should we remove support for bearer authentication, this should be removed as well
- `authorization_details_types_supported` : An empty array. To explicitly state that we don't support authorization details (as defined in OpenId4VCI)
- `dpop_signing_alg_values_supported` : An non empty array. Probably including only what is described in HAIP.
- `dpop_bound_access_tokens_required` : We should put this `false` to indicate that we don't require DPoP AT. If we decide to go only with HAIP option (by removing support for bearer access token), we should change the value to `true`

No need to signed them, for starters

dzarras:
This PR introduced support for `RFC9728: Protected Resource Metadata`.

pid-issuer currently advertised the following metadata:

* `resource`: The public URL of pid-issuer - *extracted from the configuration*
* `authorization_servers`: URLs of the Authorization Servers used by pid-issuer - *the URL of configured Authorization Server*
* ~`jwks_uri`: URL where the signing keys used by pid-issuer can be retrieved from - *URL of the metadata API used to retrieve the Public JWKS*~
* `scopes_supported`: OAuth 2.0 client scopes supported by pid-issuer - *extracted from the configured/enabled specific issuer*
* `bearer_methods_supported`: Methods supported by pid-issuer for sending an OAuth 2.0 Bearer Access Token - *always set to `bearer`*
* ~`resource_signing_alg_values_supported` JWS Algorithms used by pid-issuer to sign issued Credentials - *extracted from the configured signing key*~
* ~`authorization_details_types_supported`: Authorization Details types supported by pid-issuer - *always an empty list, as none is supported*~
* `dpop_signing_alg_values_supported`: DPoP Access Token JWS Algorithms supported by pid-issuer - *currently relays `dpop_signing_alg_values_supported` from the Authorization Server metadata*
* `dpop_bound_access_tokens_required`: Whether pid-issuer requires DPoP Access Tokens - *always set to false*

Example:
```json
{
  "resource": "http://localhost:8080",
  "authorization_servers": [
    "https://localhost/idp/realms/pid-issuer-realm"
  ],
  "scopes_supported": [
    "eu.europa.ec.eudi.pid_mso_mdoc",
    "eu.europa.ec.eudi.pid_vc_sd_jwt",
    "org.iso.18013.5.1.mDL",
    "urn:eudi:ehic:1:dc+sd-jwt",
    "urn:eu.europa.ec.eudi:learning:credential:1:dc+sd-jwt"
  ],
  "bearer_methods_supported": [
    "header"
  ],
  "dpop_signing_alg_values_supported": [
    "PS384",
    "ES384",
    "RS384",
    "ES256",
    "RS256",
    "ES512",
    "PS256",
    "PS512",
    "RS512"
  ],
  "dpop_bound_access_tokens_required": false
}
```

Closes #441

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/input/web/IssuerUi.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/input/web/IssuerUi.kt
@@ -130,6 +130,7 @@ class IssuerUi(
             )
 
         val credentialIssuerMetadata = credentialIssuer.wellKnown("openid-credential-issuer")
+        val protectedResourceMetadata = credentialIssuer.wellKnown("oauth-protected-resource")
         val authorizationServerMetadata = authorizationServer.wellKnown("oauth-authorization-server")
         val sdJwtVcIssuerMetadata = credentialIssuer.wellKnown("jwt-vc-issuer")
         val pidSdJwtVcTypeMetadata = credentialIssuer.appendPath("/type-metadata/urn:eudi:pid:1")
@@ -139,6 +140,7 @@ class IssuerUi(
 
         return mapOf(
             "credential_issuer_metadata" to credentialIssuerMetadata.externalForm,
+            "protected_resource_metadata" to protectedResourceMetadata.externalForm,
             "authorization_server_metadata" to authorizationServerMetadata.externalForm,
             "sdjwt_vc_issuer_metadata" to sdJwtVcIssuerMetadata.externalForm,
             "pid_sdjwt_vc_type_metadata" to pidSdJwtVcTypeMetadata.externalForm,
--- a/src/main/resources/i18n/messages.properties
+++ b/src/main/resources/i18n/messages.properties
@@ -20,6 +20,7 @@ useful-links=Useful links
 description=Description
 url=URL
 credential_issuer_metadata=Credential Issuer Metadata
+protected_resource_metadata=Protected Resource Metadata
 authorization_server_metadata=Authorization Server Metadata
 sdjwt_vc_issuer_metadata=SD-JWT VC Issuer Metadata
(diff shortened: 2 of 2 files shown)


src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/GetProtectedResourceMetadata.kt
```diff
@@ -0,0 +1,72 @@
...
+
+    @SerialName(RFC9728.BEARER_METHOD_BODY)
+    BODY,
+
+    @SerialName(RFC9728.BEARER_METHOD_QUERY)
+    QUERY,
+}
+
+@Serializable
+data class ProtectedResourceMetadataTO(
```
babisRoutis: Switch to NonEmptyList<T>? = null

To support arrow serialization
1) Add as api dependency `arrow-core-serialization  = { module = "io.arrow-kt:arrow-core-serialization", version.ref = "arrow" }`
2) To the 1st line of the file add

```kotlin
@file:UseSerializers(
    NonEmptyListSerializer::class,
)

```

PR #480, opened 2025-12-17
Group offered Credential Configurations by Attestation Category

### LINKED ISSUE #477: Group offered Credential Configurations by Attestation Category
A new sealed hierarchy, called `AttestationCategory` must be introduced.

The initial list of data objects, is the following:
* PID
* EU_QEEA
* EU_PUB_QEAA
* EU_EAA
* EAA

Each Credential Configuration must be updated to include an `AttestationCategory`.

Initially the supported Credential Configurations will be configured as follows:
* PID (both MSO MDoc and SD-JWT VC) -> PID
* mDL -> EU_QEEA
* EHIC (SD-JWT VC) -> EU_QEEA
* Learning Credential (SD-JWT VC) -> EAA

Given the above, the Credential Offer screen must be updated as follows:
1. Credential Configuration Ids must be grouped by Attestation Category
2. A user must be able to generate a Credential Offer only when selecting Credential Configuration Ids from the same Attestation Category

The second point affects `CreateCredentialsOffer`, but could also be implemented using Javascript in the Credential Offer screen, i.e. disable the Generate button when a user selects Credential Configuration Ids from different Attestation Categories

atsant:
Closes #477


src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/AttestationCategory.kt
```diff
@@ -0,0 +1,25 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package eu.europa.ec.eudi.pidissuer.domain
+
+@Suppress("ktlint:standard:class-naming")
+sealed interface AttestationCategory {
```
dzarras: @atsant Let's use camel case instead, and remove the suppression.

i.e.:

* `Pid`
* `EuQeaa`
* `EuPubQeaa`
* `EuEaa`
* `Eaa`

src/main/resources/templates/generate-credentials-offer-form.html
```diff
@@ -40,15 +40,24 @@ <h4 th:text="#{please-select-credentials-to-issue} + ':'">Please select the Cred
...
-                            <input class="form-check-input" type="checkbox" name="credentialIds"
-                                   value="id1" th:value="${credentialId}"
-                                   id="credentialId-1" th:id="'credentialId-' + ${iteration.index}">
-                            <label class="form-check-label"
-                                   for="credentialId-1" th:for="'credentialId-' + ${iteration.index}"
-                                   th:text="${credentialId}">credential-1</label>
+                        <div class="form-label mb-2">
+                            <h5  th:text="#{credentials}">Credentials</h5>
+                        </div>
+                        <!--/*@thymesVar id="credentialIds" type="kotlin.collections.MapsKt"*/-->
```
dzarras: @atsant The type is not correct. It should be `kotlin.collection.Map`

src/main/resources/templates/generate-credentials-offer-form.html
```diff
@@ -96,6 +105,28 @@ <h4 th:text="#{useful-links} + ':'">Useful links: </h4>
         </div>
     </div>
 </main>
+
+<script type="application/javascript">
+    (
+        function () {
+            const generateButton = document.getElementById("generateButton");
+            const checkboxes = Array.from(document.querySelectorAll(".form-check-input"));
```
dzarras: @atsant I'd suggest adding a custom class instead. For instance add the class `credential-configuration-id`.

dzarras (changes requested):

src/main/resources/templates/generate-credentials-offer-form.html
```diff
@@ -96,6 +105,28 @@ <h4 th:text="#{useful-links} + ':'">Useful links: </h4>
...
+
+            function updateGenerateEnabled() {
+                const selectedCategories = new Set(
+                    checkboxes
+                        .filter(checkbox => checkbox.checked)
+                        .map(checkbox => checkbox.dataset.category)
+                );
+
+                generateButton.disabled = selectedCategories.size > 1;
```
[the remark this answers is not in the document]
dzarras: Also come to think of it, by default the generate button should be disabled.

PR #511, opened 2026-02-26
feat: match by exact username when fetching data from keycloak

asserh:
By default, Keycloak matches usernames by substring, which can lead to unexpected or incorrect user matches. This change sets the `exact` query parameter to ensure only exact username matches.

See: https://www.keycloak.org/docs-api/latest/rest-api/index.html#_users


dzarras (approved):
@asserh Thanks for spotting this.

Kind regards.

PR #512, opened 2026-02-26
fix: make email optional in UserRepresentation to match UserInfo

asserh:
Email field is optional in the UserInfo data class; align UserRepresentation serialization to reflect the same optionality by removing the @Required annotation.


src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/pid/GetPidDataFromKeyCloak.kt
```diff
@@ -280,7 +280,7 @@ data class UserRepresentation(
     @Required val lastName: String,
     @Required val firstName: String,
     val attributes: Map<String, List<String>> = emptyMap(),
-    @Required val email: String,
+    val email: String? = null,
```
dzarras: Hi @asserh ,

Thanks for pointing this out. 

Currently in the realm definition we provide with pid-issuer, `email` is a required user attribute. Hence why it is marked `@Required` in `UserRepresentation`.

Given that [PID Rulebook](https://github.com/eu-digital-identity-wallet/eudi-doc-attestation-rulebooks-catalog/blob/main/rulebooks/pid/pid-rulebook.md#23-optional-attributes-specified-in-cir-20242977) defines `email_address` as optional, I would suggest the following:

- [x] Make `UserRepresentation.email` nullable 
- [ ] Update `kc.user.profile.config` in `org.keycloak.userprofile.UserProfileProvider`, in the realm settings in `pid-issuer-realm-realm.json`, and make `email` optional

Doing the above will ensure consistency between the User profile in KeyCloak, `UserRepresentation`, and `UserInfo`.

Thanks again.

PR #522, opened 2026-04-03
Implemented credential reuse policies

### LINKED ISSUE #505: Support ETSI TS 119 472-3 - Attestation provider Reuse Policies
ETSI TS 119 472-3 defines specific requirements on how attestation providers should advertice their policies regarding how the issued credential should be re-used when presented to relying parties.

Analysis provided [here](https://github.com/eu-digital-identity-wallet/eudi-lib-jvm-openid4vci-kt/issues/494). It describes 4 reuse policies:
- Method A - Once-only
- Method B - Limited-time
- Method C - Rotating-batch
- Method D - Per-Relying Party

As a first step kotlin issuer should support advertising reuse policy of Method A 

- [x] Issuer's configuration should be extended to support, per credential configuration, a set of intended re-use policies According to policy type its parameters should be configurable.
- [x] Include reuse policies in issuer's metadata (signed or not)
- [x] Upon receiving an issuance request validate that request aligns with the credential's reuse policies. 
     - If "limited-time" only cinfigured for credential, batch issuance must not be allowed. 
     - If a policy that allows batch issuance is configured for the credentialm the proofs sent must be aligned with the `batch_size` parameter of the policy.

gdimtsas:
Closes #505


babisRoutis:
Hi @gdimtsas 

Can you please update the PR branch?

Also, just a sanity check: In case of method B, which is used only if the issued credential is short lived and non-revocable, have we updated the issuance to not include references to a status list?

gdimtsas:
> Hi @gdimtsas
> 
> Can you please update the PR branch?
> 
> Also, just a sanity check: In case of method B, which is used only if the issued credential is short lived and non-revocable, have we updated the issuance to not include references to a status list?

No this is not implemented yet. I will go ahead and implement it.

babisRoutis:
@gdimtsas Can you please update the PR?

gdimtsas:
> @gdimtsas Can you please update the PR?

Done!

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/CredentialReusePolicy.kt
```diff
@@ -0,0 +1,171 @@
...
+    require(reissueTriggerLifetimeLeft > 0) { "'reissue_trigger_lifetime_left' must be greater than 0" }
+}
+
+/**
+ * A single ARF Annex II option in the reuse policy.
+ */
+sealed interface EudiReusePolicy {
+
+    val batchSize: Int?
+    val reissueTriggerUnused: Int?
```
babisRoutis, line 51: I would recommend using `Duration?` instead of `Int?`
But remember to check that duration is positive.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/CredentialReusePolicy.kt
```diff
@@ -0,0 +1,171 @@
...
+    PerRelyingParty("per-relying-party"),
+    ;
+
+    companion object {
+        fun fromValue(value: String): EudiReusePolicyType? = entries.firstOrNull { it.value == value }
+    }
+}
+
+private fun validateBatchSize(batchSize: Int) {
+    require(batchSize > 0) { "'batch_size' must be greater than 0" }
```
vafeini: batchSize must be >= 2

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/IssueCredential.kt
```diff
@@ -516,6 +516,29 @@ private interface Validations : Raise<IssueCredentialError> {
         ): UnresolvedCredentialRequest.ByCredentialConfigurationId =
             supportedCredentialConfigurations.firstOrNull { credentialConfigurationId == it.id }
                 ?.let { credentialConfiguration ->
+                    credentialConfiguration.credentialReusePolicy.let { reusePolicy ->
```
vafeini: This validation should not be limited to the size of proofs array. In case of `attestation` proof type or JWT with key_attestation header the actual size of keys is in the `attestated_keys` array. This is the size that needs to be 1 in the case of `LimitedTime` and  in the case of batch reuse policies compared with `effectiveBatchSize`.

But besides the above this validation is not needed in this point. The client can send as many attested keys as he wants. Its the issuers decision to use the keys as defined per reuse policy:   

- In the case of LimitedTime, only one key will be used, the rest will be discarded.
- In the case of batch policies and if client sent more keys than the ones required from the batch size, issuer should use as many keys as the batch size defines, the rest should be discarded.

vafeini (approved):

PR #541, opened 2026-05-05
Introduce `readNullableNonEmptySet` to handle optional non-empty sets.

gdimtsas:


src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerContext.kt
```diff
@@ -315,7 +315,7 @@ fun beans(clock: Clock) = BeanRegistrarDsl {
             val parameters = CredentialRequestEncryptionSupportedParameters(
                 encryptionKeys = JWKSet(key),
                 methodsSupported = encryptionMethods,
-                zipAlgorithmsSupported = env.readNonEmptySet(
+                zipAlgorithmsSupported = env.readNullableNonEmptySet(
                     "issuer.credentialRequestEncryption.zipAlgorithmsSupported",
                 ) { CompressionAlgorithm(it) },
```
dzarras: @gdimtsas Change the mapping function to something like
`algorithm -> algorithm.takeIf { it.isNotBlank() }?.let { CompressionAlgorithm(it) }`
to avoid parsing blank values

dzarras (approved):

PR #547, opened 2026-05-11
Ensure Base64 encoding uses absent padding in `MsoMdocSigner`

### LINKED ISSUE #546: PID (MSO MDoc) Credential Response returns padded base64url
### Summary

When attempting to issue a PID (MSO MDoc), the `credential` value returned is base64url-encoded CBOR, but the encoding includes padding characters (=).

This breaks compatibility with [OID4VCI 1.0](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#section-2-2), which requires base64url encoding **without padding**, as defined in [RFC7515](https://www.rfc-editor.org/rfc/rfc7515.html).

### Documentation

According to [OpenID for Verifiable Credential Issuance 1.0](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html):

[2. Terminology](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#section-2-2)
> Base64url-encoded denotes the URL-safe base64 encoding without padding defined in Section 2 of [[RFC7515](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#RFC7515)].

[8.3. Credential Endpoint - Credential Response](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#section-8.3)
> Credential Formats expressed as binary data MUST be base64url-encoded and returned as a string.

And, more specifically, 

[Appendix A.2.4 Credential Format Profiles - Mobile Documents or mdocs (ISO/IEC 18013) - Credential Response](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-credential-response-5)
> The value of the credential claim in the Credential Response MUST be a string that is the base64url-encoded representation of the CBOR-encoded IssuerSigned structure, as defined in [[ISO.18013-5](https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#ISO.18013-5)].

Therefore, the `credential` claim in the mdoc credential response MUST be a RFC7515-compliant base64url-encoded string, without padding.

### How to reproduce
Running PID issuer locally, attempt to request a credential with the identifier `eu.europa.ec.eudi.pid_mso_mdoc`.

Following the OID4VCI issuance flow, the Credential Response object that is received contains

```
{
    "credentials": [
        {
            "credential": "ompuYW1lU3BhY2Vz...aj2kHOtwayIm2S4="
        }
    ]
}
```
You can see the last character is a padding symbol `=`.
This causes _zod's_ `z.base64url()` validation to fail, which also affects the _owf/mdoc-ts_ library, as it uses this validator.

It is worth noting that the Mobile Driving Licence (credential with identifier `org.iso.18013.5.1.mDL`) does not suffer from this issue, as it does not include padding, but it might be due to a lucky input length.

### Suggested fix
_MsoMdocSigner.sign()_ function uses `Base64.UrlSafe.encode(...)`.
Kotlin standard library's Base64 implementation provides the option not to use padding [PaddingOption.ABSENT](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.io.encoding/-base64/-padding-option/-a-b-s-e-n-t/).
Replacing the above usage with
`Base64.UrlSafe.withPadding(Base64.PaddingOption.ABSENT).encode(...)`
might fix this issue.

atsant:
Closes #546

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/msomdoc/MsoMdocSigner.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/msomdoc/MsoMdocSigner.kt
@@ -35,7 +35,6 @@ import id.walt.mdoc.mso.DeviceKeyInfo
 import id.walt.mdoc.mso.MSO
 import id.walt.mdoc.mso.ValidityInfo
 import kotlinx.datetime.toDeprecatedInstant
-import kotlin.io.encoding.Base64
 import kotlin.time.Instant
 
 internal class MsoMdocSigner<in Credential>(
--- a/src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/msomdoc/MsoMdocSigner.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/msomdoc/MsoMdocSigner.kt
@@ -18,6 +18,7 @@ package eu.europa.ec.eudi.pidissuer.adapter.out.msomdoc
 import COSE.OneKey
 import com.nimbusds.jose.jwk.ECKey
 import eu.europa.ec.eudi.pidissuer.adapter.out.IssuerSigningKey
+import eu.europa.ec.eudi.pidissuer.adapter.out.base64.base64UrlSafeNoPadding
 import eu.europa.ec.eudi.pidissuer.adapter.out.cryptoProvider
 import eu.europa.ec.eudi.pidissuer.domain.MsoDocType
 import eu.europa.ec.eudi.pidissuer.domain.StatusListToken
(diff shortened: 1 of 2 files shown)


src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/msomdoc/MsoMdocSigner.kt
```diff
@@ -64,7 +64,7 @@ internal class MsoMdocSigner<in Credential>(
         val mdoc = MDocBuilder(docType)
             .apply { usage(credential) }
             .sign(validityInfo, deviceKeyInfo, statusListToken, issuerCryptoProvider, issuerSigningKey.key.keyID)
-        return Base64.UrlSafe.encode(mdoc.issuerSigned.toMapElement().toCBOR())
+        return Base64.UrlSafe.withPadding(Base64.PaddingOption.ABSENT).encode(mdoc.issuerSigned.toMapElement().toCBOR())
```
dzarras: @atsant Please introduce an `internal` `Base64` instance named `base64UrlSafeNoPadding` in a file named `Base64Utils.kt` in the adapters package, so that it can be reused wherever it might be needed.

dzarras (approved):

PR #548, opened 2026-05-11
Support mapping of `personal_administrative_number` from user data in Keycloak

### LINKED ISSUE #545: Support mapping of personal administrative number from user data in Keycloak
[The PID rulebook](https://github.com/eu-digital-identity-wallet/eudi-doc-attestation-rulebooks-catalog/blob/main/rulebooks/pid/pid-rulebook.md) describes the personal administrative number in the following way:
> A value assigned to the natural person that is unique among all personal administrative numbers issued by the provider of person identification data. Where Member States opt to include this attribute, they shall describe in their electronic identification schemes under which the person identification data is issued, the policy that they apply to the values of this attribute, including, where applicable, specific conditions for the processing of this value.

The current behavior of the PID issuer is to populate this attribute with a random UUID unique for each issued PID. To allow users of the PID-issuer better control over this value and to enable more realistic data we suggest to change the logic so that the PID issuer populates the `personal_administrative_number` from a corresponding attribute in the user data coming from Keycloak. This way, we can create users in Keycloak with unique and realistic values for this datapoint.

In order not to disrupt existing users of the PID issuer, we suggest that we fallback to the current behavior of using a random UUID when there is no value coming from Keycloak.

A working proof-of-concept is already available in diggsweden/eudi-srv-pid-issuer#2. However, we should probably add automated tests covering the changes and possibly also update the documentation somehow. If desired, we should also update the Keycloak realm configuration we have in version control so that it supports this data point. A working example is available in diggsweden/wallet-ecosystem#169.

atsant:
Now the field  `personal_administrative_number` is not automatically generated.

Closes #545


src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/pid/GetPidDataFromKeyCloak.kt
```diff
@@ -270,6 +270,9 @@ class GetPidDataFromKeyCloak(
             sex = userInfo.gender?.let { IsoGender(it) },
             emailAddress = userInfo.email,
             mobilePhoneNumber = null,
+            personalAdministrativeNumber = userInfo.personalAdministrativeNumber?.let {
```
dzarras, line 273: @atsant On seconds thoughts, I'd suggest we remove the fallback value. Per PID Rulebook v1.5 `personal_administrative_number` is an optional PID attribute. No need to have a default.

atsant: Yeah, I noticed that as well. Currently, it's an optional field but in reality it isn't since it is always populated.

dzarras (approved):

PR #549, opened 2026-05-12
Remove support for JWT proofs without a key attestation

gdimtsas:
- Remove support for JWT proofs without a key attestation. 
- Limit the maximum supported number of proofs to 1.


src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/ValidateJwtProof.kt
dzarras: @gdimtsas We should `KeyAttestationRequirement` to a data class that contains all the members of `KeyAttestationRequirement.Required`. It doesn't make sense to keep it as a `sealed interface` with two members given we no longer support non-attested keys.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/ValidateJwtProof.kt
```diff
@@ -108,14 +103,11 @@ private suspend fun algorithmAndCredentialKey(
     }
 
     val key = when {
-        kid != null && jwk == null && keyAttestation == null && x5c.isNullOrEmpty() -> CredentialKey.DIDUrl(kid).getOrThrow()
-        kid == null && jwk != null && keyAttestation == null && x5c.isNullOrEmpty() -> CredentialKey.Jwk(jwk)
-        kid == null && jwk == null && keyAttestation == null && !x5c.isNullOrEmpty() -> CredentialKey.X5c.parseDer(x5c).getOrThrow()
-        jwk == null && keyAttestation != null && x5c.isNullOrEmpty() -> {
+        keyAttestation != null -> {
```
dzarras: @gdimtsas `keyAttestation` cannot be null.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/ValidateJwtProof.kt
```diff
@@ -165,25 +157,11 @@ private suspend fun CredentialKey.ensureCompatibleWithAlgorithm(algorithm: JWSAl
     }
 
     when (this) {
-        is CredentialKey.DIDUrl -> jwk.ensureCompatibleWith(algorithm)
-        is CredentialKey.Jwk -> value.ensureCompatibleWith(algorithm)
-
         is CredentialKey.AttestedKeys -> {
             val signingJWK = keys.signingKeyOf(signedJwt)
```
dzarras: @gdimtsas `signingJWK` must be the first of the attested keys. The logic that was previously in `keys.signingKeyOf(signedJwt)` is no longer valid.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/ValidateJwtProof.kt
dzarras: @gdimtsas Same here. `signingJWK` must be the first of the attested keys.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/ValidateProofs.kt
```diff
@@ -43,33 +44,28 @@ internal class ValidateProofs(
...
-                when (it) {
+            val credentialKeysAndCNonces =
+                when (unvalidatedProof) {
                     is UnvalidatedProof.Jwt ->
-                        validateJwtProof(it, credentialConfiguration, at).bind()
+                        validateJwtProof(unvalidatedProof, credentialConfiguration, at).bind()
                     is UnvalidatedProof.Attestation ->
-                        validateAttestationProof(it, credentialConfiguration, at).bind()
+                        validateAttestationProof(unvalidatedProof, credentialConfiguration, at).bind()
                     is UnvalidatedProof.DiVp -> raise(InvalidProof("Supporting only JWT proof"))
```
dzarras: @gdimtsas Let's remove `UnvalidatedProof.DiVp` and keep only the members that are valid in our case. i.e. JWT Proofs and Attestation Proofs.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/ValidateProofs.kt
```diff
@@ -43,33 +44,28 @@ internal class ValidateProofs(
...
                 }
-            }
 
-            val cnonces = credentialKeysAndCNonces.map { it.second }.toNonEmptyListOrNull()
-            checkNotNull(cnonces)
-            ensure(verifyNonce(cnonces, at)) {
+            val cNonces = credentialKeysAndCNonces.second
```
dzarras: @gdimtsas Given we no longer accept multiple proofs, we expect a single cnonce. Let's rename the variables accordingly.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/DefaultExtractJwkFromCredentialKey.kt
```diff
@@ -16,16 +16,11 @@
...
-import com.nimbusds.jose.jwk.JWK
 import eu.europa.ec.eudi.pidissuer.domain.CredentialKey
 
 val DefaultExtractJwkFromCredentialKey = ExtractJwkFromCredentialKey { key ->
     Either.catch {
         when (key) {
-            is CredentialKey.Jwk -> nonEmptyListOf(key.value)
-            is CredentialKey.X5c -> nonEmptyListOf(JWK.parse(key.certificate))
-            is CredentialKey.DIDUrl -> nonEmptyListOf(key.jwk)
             is CredentialKey.AttestedKeys -> key.keys
```
dzarras: @gdimtsas I'd suggest we convert `CredentialKey` to a data class that contains the members of `CredentialKey.AttestedKeys`. I don't think it makes sense to keep it as a sealed hierarchy.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/CredentialRequest.kt
```diff
@@ -58,45 +54,6 @@ sealed interface UnvalidatedProof {
  */
 sealed interface CredentialKey {
```
dzarras: @gdimtsas Let's convert it to a simple data class instead.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/IssueCredential.kt
dzarras: @gdimtsas Given the check at line 504, I'd remove this. It's redundant.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/IssueCredential.kt
```diff
@@ -501,12 +501,13 @@ private interface Validations : Raise<IssueCredentialError> {
 
                     val proofs = (jwtProofs.orEmpty() + diVpProofs.orEmpty() + attestations.orEmpty()).toNonEmptyListOrNull()
                     ensureNotNull(proofs) { MissingProof }
+                    ensure(proofs.size == 1) {
+                        InvalidProof("You can provide at most 1 proof")
+                    }
+                    proofs.firstOrNull()
```
dzarras: @gdimtsas Please change this to `proofs.first()`. We don't support any non-device-bound attestations, so proof is always required.

dzarras (changes requested):

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/ValidateJwtProof.kt
gdimtsas: Done

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/ValidateProofs.kt
gdimtsas: Done

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/CredentialRequest.kt
gdimtsas: Done!

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/DefaultExtractJwkFromCredentialKey.kt
gdimtsas: Done! I actually converted it to `value class CredentialKeys(val value: NonEmptyList<JWK>)`

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/IssueCredential.kt
gdimtsas: Done

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/ValidateProofs.kt
gdimtsas: Done

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/IssueCredential.kt
gdimtsas: Done

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/ValidateJwtProof.kt
gdimtsas: Done

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/ValidateJwtProof.kt
gdimtsas: Done

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/CredentialRequest.kt
```diff
@@ -56,62 +47,20 @@ sealed interface UnvalidatedProof {
...
-        companion object {
-            private fun JWK.isPublicAsymmetricKey() {
-                require(!isPrivate) {
-                    "Private key provided while expecting a public key."
-                }
-                require(this is AsymmetricJWK) {
-                    "Symmetric key provided. Must be an asymmetric key."
-                }
+    companion object {
+        private fun JWK.isPublicAsymmetricKey() {
```
dzarras: @gdimtsas Given the function name, I would expect it to return a `Boolean`. Better rename it to `ensureIsPublicAsymmetricKey()`

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/ValidateJwtProof.kt
```diff
@@ -198,16 +159,9 @@ private suspend fun keySelector(
...
-        is CredentialKey.AttestedKeys -> {
-            val signingJWK = credentialKey.keys.signingKeyOf(signedJwt)
-            requireNotNull(signingJWK) { "Key attestation does not contain a key that verifies the jwt proof signature" }
-            signingJWK.keySelector(algorithm)
-        }
-        is CredentialKey.DIDUrl -> credentialKey.jwk.keySelector(algorithm)
-        is CredentialKey.Jwk -> credentialKey.value.keySelector(algorithm)
-        is CredentialKey.X5c -> SingleKeyJWSKeySelector(algorithm, credentialKey.certificate.publicKey)
-    }
+    val signingJWK = credentialKeys.value.signingKeyOf(signedJwt)
```
dzarras: @gdimtsas In this context we create a `KeySelector` for a `JWTProcessor` to check the signature of the `SignedJWT`, yet `signingKeyOf` already checks internally the signature. I find this redundant and confusing. I would remove the function `signingKeyOf` all-together, set `signingJwk` as `credentialsKeys.value.first()` instead.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/ValidateJwtProof.kt
```diff
@@ -164,32 +143,14 @@ private suspend fun CredentialKey.ensureCompatibleWithAlgorithm(algorithm: JWSAl
...
-                when (certificate.publicKey) {
-                    is ECPublicKey -> ECDSASigner.SUPPORTED_ALGORITHMS
-                    else -> error("Certificate key not supported")
-                }
-            require(algorithm in supportedAlgorithms) {
-                "certificate algorithm '${certificate.publicKey.algorithm}' is not compatible with signing algorithm '${algorithm.name}'"
-            }
-        }
-    }
+    val signingJWK = value.signingKeyOf(signedJwt)
```
dzarras: @gdimtsas I would remove `signingKeyOf` all-together. We know `signingJWK` must be `value.first()`. Additionally in this context we want to check whether the `signingJwk` is compatible with the signing algorithm of the `SignedJWT`, yet `signingKeyOf` check the signature internally and returns a `JWK` only when the signature validation succeeds, i.e. the `JWK` is indeed compatible with the signature algorithm.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/CredentialConfiguration.kt
```diff
@@ -41,14 +41,16 @@ value class AttackPotentialResistance(val value: String) {
...
-    data class Required(
-        val keyStorage: NonEmptySet<AttackPotentialResistance>?,
-        val userAuthentication: NonEmptySet<AttackPotentialResistance>?,
-    ) : KeyAttestationRequirement
+data class KeyAttestationRequirement(
+    val keyStorage: NonEmptySet<AttackPotentialResistance>?,
+    val userAuthentication: NonEmptySet<AttackPotentialResistance>?,
+) {
+    companion object {
+        fun plain(): KeyAttestationRequirement = KeyAttestationRequirement(
```
dzarras: @gdimtsas Per TS3:

> For a KA about a WSCD, the key_storage and user_authentication attributes shall be iso_18045_high as the WSCD by definition must ensure LoA High.

Instead of using `plain()` as a default, we should probably introduce a `ts3()` factory and use this as a default value instead.

@babisRoutis, @vafeini In my opinion, it also makes sense to remove the configuration options for key attestation requirement, and hardcode the values mandated by TS3. What do you think?

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/ValidateProofs.kt
```diff
@@ -39,37 +38,28 @@ internal class ValidateProofs(
...
                 InvalidNonce("CNonce is not valid")
             }
 
-            val jwks = credentialKeysAndCNonces.map {
-                extractJwkFromCredentialKey(it.first).getOrElse { error ->
-                    raise(InvalidProof("Unable to extract JWK from CredentialKey", error))
-                }
-            }.flatten()
-                .distinct()
+            val jwks = credentialKeysAndCNonce.first.value.distinct()
```
dzarras: @gdimtsas I would remove `.distinct()` and add a requirement in the `init` block of `CredentialKeys` to ensure we have distinct elements.

dzarras (changes requested):

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/jose/ValidateJwtProof.kt
gdimtsas: You're right, it is redundant. I will remove it

gdimtsas: @dzarras The only benefit i see in keeping `signingKeyOf` is that it produces a clear error: `Invalid proof JWT: The first key in the key attestation does not verify the jwt proof signature`. 
The JWTProcessor produces a more cryptic `Invalid proof JWT: Signed JWT rejected: Invalid signature`.

But I'm thinking it is better to remove it and simplify.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/CredentialRequest.kt
gdimtsas: Done

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/CredentialConfiguration.kt
gdimtsas: I removed the configuration options. Since TS3 mandates the highest level of security, it doesn't leave anything open to configuration.

dzarras (changes requested):
@gdimtsas Code-wise no further comments from my side. Please update README.md and remove any environment variables that are no longer applicable/used. We should also include a notice that pid-issuer requires Key Attestations with the requirements set by TS3.

dzarras (approved):

Issue #529, opened 2026-04-16
PID\Attestation issuance adaptations to TS3 v1.5


vafeini:
**KA keys**

> The attested_keys element of an key_attestation SHALL contain at least one key, but MAY contain multiple keys (in order to support batch issuance). The number of keys in the KA sent to a PID Provider or Attestation Provider SHOULD NOT exceed the maximum batch size specified by that Attestation Provider or PID Provider in the Credential Issuer metadata for this type of credential; see ETSI TS 119 472-3, parameter credential_configurations_supported.credential_metadata.credential_reuse_policy.options.batch_size.
>NOTE: If the number of keys exceeds the batch size, the PID Provider or Attestation Provider will not use all keys in the KA to bind PIDs or attestations to. Since a Wallet Unit uses a KA at most once, this implies that the unused keys will never be used. It is up to the Wallet Provider to devise a strategy for minimizing the number of keys that will remain unused, if needed.

- [ ] If the number of keys in ΚΑ exceeds the batch size defined in issuer's re-use policies, issuer will not use all keys in the ΚΑ.

**Technical Vallidity Period**

TS3 specification mentions:

>The technical validity period of a PID SHALL end before the client_status.exp of the WIA and the key_storage_status.exp of the KA shown to the PID Provider in the issuance process.
> Other Attestation Providers (i.e., non-PID Providers) MAY choose a technical validity period of the attestations they issue independently of the client_status.exp of the WIA and the key_storage_status.exp of the KA received during issuance.

- [ ] For simplicity of implementation we will follow the same approach for all attestations. Technical validity period of attestations should be before the minimum of `client_status.exp` and `key_storage_status.exp`.

**Monitor status of KA and WIA**

 - [ ] Extract `client_status` from access token and associate with scope of issuance (attestation issued in this context)
 - [ ] Associate issued attestations instances with received `client_status` and `key_storage_status`
 - [ ] Check the revocation status of the ΚΑ and WIA received in relation to issuance at least once every 24 hours 
 - [ ] If the ΚΑ or WIA is revoked, then the associated attestations must be revoked from issuer.

**Issuer Persistence** 

To support the above issuer must be extended to support persistence of issued attestations' metadata. For the time being we have an in memory storage. This is not enough any more, to support the monitoring of WIA and KA statuses. A document non-relational database seems a good candidate for persistence.

[...]

PR #552, opened 2026-05-15
Enforce credentials expiry date based on client status and key storage status expiration dates

### LINKED ISSUE #530: Issuers metadata updates to align with TS3 v1.5
TS3 in section 2.4 defined that issuer metadata should be extended with:

> The following field is to be placed within the key_attestations_required object as specified in https://github.com/eu-digital-identity-wallet/eudi-doc-standards-and-technical-specifications/issues/3:
> - preferred_key_storage_status_period: OPTIONAL. An integer specifying a PID or Attestation Provider's preference for the remaining status maintenance period (key_storage_status.exp minus current time) of the KA it receives during issuance, in seconds. See [Section 2.4.2](https://github.com/eu-digital-identity-wallet/eudi-doc-standards-and-technical-specifications/blob/main/docs/technical-specifications/ts3-wallet-unit-attestation.md#242-wallet-provider-responsibilities) for how the Wallet Unit and the Wallet Provider must handle this parameter.

- [ ] For key-bound attestations include attribute `preferred_key_storage_status_period` in `key_attestations_required` element of `jwt` and `attestation` proof types of `proof_types_supported` array. Specifically for PIDs the minimum value should be **one month**.
   
TS3 in section 2.2.2.2 mentions:

> A PID Provider or an Attestation Provider issuing device-bound attestations SHALL indicate in the proof_types_supported parameter in its Issuer Credential Metadata (see https://github.com/eu-digital-identity-wallet/eudi-doc-standards-and-technical-specifications/issues/3) that it supports both the jwt and the attestation proof types for KAs, both of which SHALL include the key_attestation_required object.

- [ ] For key-bound credentials (have `proof_types_supported` in credential configuration) both `jwt` and `attestation` proof types must be advertised and include `key_attestations_required` object. 

- [ ] Elements of `proof_types_supported` array with `key_attestations_required` specified must have their `proof_signing_alg_values_supported` array containing algorithms of the following set: ES256, ES384, ES512  

ΙΑ 2024/2979 mentions the following in Annex Ib:

> TR_KA-4.1: Where a credential issuer issues non-device-bound credentials, it shall omit the "proof_types_supported" and the "cryptographic_binding_methods_supported" parameters in the Credential Issuer Metadata.

- [ ] Either both `cryptographic_binding_methods_supported` and `proof_types_supported` should be present in a credential configuration or both missing

gdimtsas:
Partial implementation of #529, and #530


src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/credential/EnsureCredentialMinExpiration.kt
```diff
@@ -0,0 +1,39 @@
...
+
+import arrow.core.Either
+import arrow.core.raise.either
+import arrow.core.raise.ensure
+import eu.europa.ec.eudi.pidissuer.domain.ClientStatus
+import eu.europa.ec.eudi.pidissuer.domain.KeyStorageStatus
+import eu.europa.ec.eudi.pidissuer.port.input.IssueCredentialError.Unexpected
+import kotlin.time.Instant
+
+object EnsureCredentialMinExpiration {
```
dzarras: @gdimtsas I'd probably define this as an extension function over `Raise<IssueCredentialError.Unexpected>` that returns `Unit`, instead. Not really sure it's worth using an `object`.

Also the name is misleading. This check ensures that the client status expiration, and key storage status expiration satisfy/meet/fulfill the credential expiration.

gdimtsas: @dzarras converted to `fun Raise<Unexpected>.ensureCredentialExpirationIsMet` and moved it to `IssueCredential` instead of its own file

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/CredentialIssuerMetaData.kt
```diff
@@ -276,6 +276,13 @@ sealed interface BatchCredentialIssuance {
 @JvmInline
 value class PreferredClientStatusPeriod(val value: Duration) {
     init {
-        require(value >= 31.days) { "Preferred client status period must be greater than 31 days" }
+        require(value >= 31.days) { "Preferred client status period must be at least 31 days" }
+    }
+}
+
+@JvmInline
+value class PreferredKeyStorageStatus(val value: Duration) {
```
dzarras: @gdimtsas Please rename the value class to `PreferredKeyStorageStatusPeriod` instead.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/CredentialConfiguration.kt
```diff
@@ -46,11 +47,13 @@ value class AttackPotentialResistance(val value: String) {
 data class KeyAttestationRequirement(
     val keyStorage: NonEmptySet<AttackPotentialResistance>?,
     val userAuthentication: NonEmptySet<AttackPotentialResistance>?,
+    val preferredKeyStorageStatusPeriod: PreferredKeyStorageStatus,
 ) {
     companion object {
         fun ts3(): KeyAttestationRequirement = KeyAttestationRequirement(
```
dzarras: @gdimtsas Please make `preferredKeyStorageStatusPeriod` a parameter that has a default value of `31 days`. This parameter must be configured using the `validity` of each enabled Credential Configuration. If for instance we configure EHIC with a validity of 90 days, we must advertise this value as `preferred_key_storage_status_period` and not `31 days`.

Please update `IssuerContext.kt` accordingly.
We might have to update the default validity values of Credential Configurations as well. I recall some use 30 days as default, which is not compatible with `preferred_key_storage_status_period`.

dzarras (changes requested):

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/CredentialConfiguration.kt
dzarras: Update README.md if needed as well.

dzarras (approved):

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/CredentialIssuerMetaData.kt
gdimtsas: Renamed

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/CredentialConfiguration.kt
gdimtsas: Done!

PR #553, opened 2026-05-18
Update Issuer metadata to align with TS3 v1.5

### LINKED ISSUE #528: Update Issuer metadata to align with TS3 v1.5
[TS3 v1.5](https://github.com/eu-digital-identity-wallet/eudi-doc-standards-and-technical-specifications/blob/main/docs/technical-specifications/ts3-wallet-unit-attestation.md) defines attribute `preferred_client_status_period` for issuer metadata:

> preferred_client_status_period: OPTIONAL. An integer specifying a PID or Attestation Provider's preference for the remaining status maintenance period (client_status.exp minus current time) of the WIA it receives during issuance, in seconds. This field is placed at the top level of the Credential Issuer Metadata (not within key_attestations_required). See [Section 2.4.2](https://github.com/eu-digital-identity-wallet/eudi-doc-standards-and-technical-specifications/blob/main/docs/technical-specifications/ts3-wallet-unit-attestation.md#242-wallet-provider-responsibilities) for how the Wallet Unit and Wallet Provider must handle this parameter.

Also regarding the value of this attribute TS3 mentions

>A Wallet Provider SHALL ensure that a Wallet Unit can always present WIAs and KAs whose client_status.exp and key_storage_status.exp (respectively) are at least **31 days** in the future at the time of presentation to a PID Provider or Attestation Provider. This guarantees that PID Providers can rely on revocation chaining without being forced to issue short-lived PIDs.

- [ ] Update issuer metadata to advertice this property. Minimum value should be 31 days to accomodate the requirement for PIDs

atsant:
Close #528


src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/pid/IssueSdJwtVcPid.kt
dzarras, line 199: Let's remove `calculateExpiresAt` given we introduce `validity`.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/CredentialIssuerMetaData.kt
```diff
@@ -225,6 +227,7 @@ data class CredentialIssuerMetaData(
     val credentialResponseEncryption: CredentialResponseEncryption,
     val display: List<CredentialIssuerDisplay> = emptyList(),
     val specificCredentialIssuers: List<IssueSpecificCredential>,
```
dzarras: Let's convert `specificCredentialIssuers` to a `NonEmptyList`. It doesn't make sense to start pid-issuer when no Credential Issuers have been configured/enabled.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/CredentialIssuerMetaData.kt
```diff
@@ -225,6 +227,7 @@ data class CredentialIssuerMetaData(
...
+    val preferredClientStatusPeriod: Duration,
```
dzarras: Please introduce a value class for Preferred Client Status Period, which includes all the invariants you have to check in its `init` block.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerContext.kt
```diff
@@ -1260,8 +1271,14 @@ data class KeycloakConfigurationProperties(
  */
 @ConfigurationProperties("issuer.metadata")
 internal data class IssuerMetadataProperties(
+    private val preferredClientStatusPeriodConfig: String,
```
dzarras: Use `java.time.Duration` here instead to allow Spring to convert the string value automatically. You can then convert to `kotlin.time.Duration` using the `toKotlinDuration()` extension.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerContext.kt
```diff
@@ -1260,8 +1271,14 @@ data class KeycloakConfigurationProperties(
...
     val display: List<DisplayProperties> = emptyList(),
 ) {
+    val preferredClientStatusPeriod: Duration =
+        Duration.parse(preferredClientStatusPeriodConfig).also {
+            require(it.isPositive()) { "preferredClientStatusPeriod must be positive" }
```
dzarras: No need for these checks to be here. Introduce them in the new value class.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerContext.kt
```diff
@@ -1329,3 +1346,10 @@ private fun toEnvironmentVariable(property: String): String {
         .replace("-", "")
         .uppercase()
 }
+
+private fun ensureValidPreferredClientStatusPeriod(preferredClientStatusPeriod: Duration, maxValidityOfIssuedCredentials: Duration) {
```
dzarras: I'd suggest inlining this check. It's applicable only in a single place. This feels like an invariant that belongs to the `init` block of `CredentialIssuerMetaData`

README.md
```diff
@@ -633,6 +633,10 @@ Default value: N/A
 
 ### Metadata configuration
 
+Variable: `ISSUER_METADATA_PREFERRED_CLIENT_STATUS_PERIOD_CONFIG`
```
dzarras: Let's rename this to `ISSUER_METADATA_PREFERRED_CLIENT_STATUS_PERIOD` instead.

dzarras (changes requested):

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/CredentialIssuerMetaData.kt
```diff
@@ -261,3 +272,10 @@ sealed interface BatchCredentialIssuance {
         }
     }
 }
+
+@JvmInline
+value class ClientStatusPeriod(val value: Duration) {
```
dzarras: Please rename the class to `PreferredClientStatusPeriod` instead.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/CredentialIssuerMetaData.kt
```diff
@@ -261,3 +272,10 @@ sealed interface BatchCredentialIssuance {
...
+    init {
+        require(value >= 31.days) { "Client status period must be greater than 31 days" }
```
dzarras: Please update the error message. It's the `preferred client status period`

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/GetCredentialIssuerMetaData.kt
```diff
@@ -61,6 +61,8 @@ data class CredentialIssuerMetaDataTO(
     val credentialConfigurationsSupported: JsonObject,
     @SerialName("openid4vci_version")
     val openid4VciVersion: String? = null,
+    @Required @SerialName("preferred_client_status_period")
```
dzarras: Please introduce a constant in the relevant spec instead.

README.md
```diff
@@ -633,6 +633,10 @@ Default value: N/A
 
 ### Metadata configuration
 
+Variable: `ISSUER_METADATA_PREFERRED_CLIENT_STATUS_PERIOD`  
+Description: Duration of the preferred client status period. Can not be less than 31 days.
```
dzarras: let's update the first part of the description to `Preferred Client Status maintenance period"

dzarras (changes requested):

README.md
atsant: @dzarras Should we drop that the value can not be less than 31 days as well?

dzarras: No, we have to document this.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/IssueCredential.kt
```diff
@@ -113,10 +113,15 @@ sealed interface IssueCredentialError {
     data class InvalidNonce(val msg: String, val cause: Throwable? = null) : IssueCredentialError
 
     /**
-     * Indicates a credential request contained contains an invalid 'credential_response_encryption_alg'.
+     * Indicates a credential request contains an invalid 'credential_response_encryption_alg'.
      */
     data class InvalidEncryptionParameters(val error: Throwable) : IssueCredentialError
 
+    /**
+     * Indicates a credential request contains a 'client_status.exp` value after a preferred client status period'.
```
dzarras: @atsant given the error is generic, let's not make the description so specific. we might reuse this error in the furture. for instance the client status can be invalid because it has been revoked. simply mention in the description that the client status is not valid. finally, the client status is not part of the credential request.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/IssueCredential.kt
```diff
@@ -364,6 +374,12 @@ private class Services(
                     credentialIssuerMetadata.batchCredentialIssuance,
                     credentialIssuerMetadata.credentialConfigurationsSupported,
                 )
+
+            val preferredClientStatusPeriod = credentialIssuerMetadata.preferredClientStatusPeriod.value
+            ensure((authorizationContext.clientStatus.expiresAt - clock.now()) >= preferredClientStatusPeriod) {
+                InvalidClientStatus("Client Status is before preferred client status period")
```
dzarras: @atsant the error message doesn't reflect what you are checking here. if the invariant is false, it means the client status expires before the preferred client status period.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/IssueCredential.kt
```diff
@@ -173,6 +178,9 @@ enum class CredentialErrorTypeTo {
 
     @SerialName("invalid_encryption_parameters")
     INVALID_ENCRYPTION_PARAMETERS,
+
+    @SerialName("invalid_client_status")
```
dzarras: @atsant The error codes in this enum are per OpenId4VCI. We cannot introduce an non-existing error code. Our domain errors do not map one-to-one to OpenId4VCI errors. For `InvalidClientStatus` I would probably map it to `credential_request_denied`. If not, let use the more generic error code `invalid_request` instead.

Reference: https://openid.net/specs/openid-4-verifiable-credential-issuance-1_0.html#name-credential-request-errors

dzarras (changes requested):

dzarras:
@gdimtsas Could you also have a quick look?

dzarras (approved):

gdimtsas (approved):

PR #561, opened 2026-05-27
Use the latest ABCA extension snapshot in KeyCloak docker image

dzarras:
This PR adds the ability to resolve SNAPSHOT version of the ABCA extension for KeyCloak.


docker-compose/docker-compose.yaml
```diff
@@ -9,7 +9,9 @@ services:
     build:
       context: ./keycloak
       dockerfile: Dockerfile
-    image: keycloak-abca:0.1.0
+      args:
+        VERSION: 0.2.0-SNAPSHOT
+    image: keycloak-abca:0.2.0-SNAPSHOT
```
gdimtsas: A minor comment: To avoid repeating the version twice, you could add a .env file containing a var like `APP_VERSION=0.2.0-SNAPSHOT` and then reference it from the docker-compose.yaml like this:
```yaml
...
services:
  keycloak:
    build:
      context: ./keycloak
      dockerfile: Dockerfile
      args:
        VERSION: ${APP_VERSION}
    image: keycloak-abca:${APP_VERSION}
...
```

docker-compose/keycloak/Dockerfile
```diff
@@ -1,8 +1,20 @@
...
+    else \
+      REPOSITORY="https://repo.maven.apache.org/maven2/"; \
+    fi && \
+    mvn dependency:get \
+      -DremoteRepositories=${REPOSITORY} \
+      -Dartifact="eu.europa.ec.eudi:abca-keycloak-ext:${VERSION}" \
+      -Dtransitive=false
+RUN mvn dependency:copy \
+      -Dartifact="eu.europa.ec.eudi:abca-keycloak-ext:${VERSION}" \
+      -DoutputDirectory=/tmp
```
atsant, line 14: Minor comment: I would suggest to export those (especially the 1st) to a script file ex: `artifactRetrieve.sh`

atsant (dismissed):
@dzarras only a minor comment with little to no importance.

gdimtsas (approved):

docker-compose/docker-compose.yaml
dzarras: Nice. Didn't know docker compose supported .env files. Added the .env file as requested.

docker-compose/keycloak/Dockerfile
dzarras: For the time being I'll leave it as it is. Even in the shell script the syntax wouldn't improve.
Finally, given we're talking about two commands in essence, I think it's better to have them in context in the Dockerfile, than having to check an external file.

PR #563, opened 2026-06-02
Upgrade Keycloak and Nimbus

atsant:
- Updated Keycloak setup and realm configuration 
- Updated DPoP configuration by removing parameters and switching to `InMemoryDPoPSingleUseChecker`.
- Upgraded the Keycloak base image to 26.6.2-2.
- Changed Keycloak health check handling and enabled health-related configuration.
- Bumped Nimbus JOSE JWT and OAuth2/OIDC SDK dependencies.


docker-compose/keycloak/conf/keycloak.conf
```diff
@@ -2,6 +2,9 @@ bootstrap-admin-username=admin
 bootstrap-admin-password=password
 
 http-enabled=true
+health-enabled=true
```
dzarras: @atsant please move this after `proxy-headers` to keep http related options grouped

docker-compose/keycloak/conf/keycloak.conf
```diff
@@ -2,6 +2,9 @@ bootstrap-admin-username=admin
...
+http-management-health-enabled=true
+http-management-scheme=http
```
dzarras: @atsant Please do not update the management interface protocol to `http`. Let it use `https`, same as the main interface.

docker-compose/keycloak/extra/health-check.sh
```diff
@@ -1,11 +1,19 @@
 #!/bin/bash
```
dzarras: @atsant Could you check how we can incorporate the command available here: https://www.keycloak.org/observability/health#_healthcheck in the healthcheck section of docker-compose and remove this file all together?

docker-compose/keycloak/realms/pid-issuer-realm-realm.json
```diff
@@ -787,7 +788,15 @@
         "post.logout.redirect.uris": "+",
         "oauth2.device.authorization.grant.enabled": "false",
         "display.on.consent.screen": "false",
-        "backchannel.logout.revoke.offline.tokens": "false"
+        "allow.token.introspection.without.audience.check": "true",
+        "backchannel.logout.revoke.offline.tokens": "false",
+        "standard.token.exchange.enabled": false,
+        "oauth2.jwt.authorization.grant.enabled": false,
+        "dpop.bound.access.tokens": "false",
+        "login_theme": "",
```
dzarras: @atsant Are `login_theme`, `consent.screen.text`, `backchannel.logout.ur, required?

src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerContext.kt
```diff
@@ -814,11 +814,9 @@ fun beans(clock: Clock) = BeanRegistrarDsl {
         algorithms?.let {
             log.info("DPoP support will be enabled. Supported algorithms: $it")
 
-            val proofMaxAge = env.duration("issuer.dpop.proof-max-age") ?: 1.minutes
-            val cachePurgeInterval = env.duration("issuer.dpop.cache-purge-interval") ?: 10.minutes
             val realm = env.getProperty("issuer.dpop.realm")?.takeIf { it.isNotBlank() }
 
-            registerBean { DPoPConfigurationProperties(it, proofMaxAge, cachePurgeInterval, realm) }
+            registerBean { DPoPConfigurationProperties(it,   realm) }
```
dzarras: @atsant Please check the formatting.

dzarras (changes requested):

docker-compose/keycloak/conf/keycloak.conf
atsant: With https://github.com/eu-digital-identity-wallet/eudi-srv-pid-issuer/pull/563#discussion_r3340602255 , we are not able to set the `http-management-scheme` to https because the healthcheck relies on `http` call. 
Moreover, since we are exposing everything via HAProxy http will never be publicly available but only available inside the container.

atsant: @dzarras Also, the old healthcheck script has been removed.

dzarras: The healthcheck is a command that run **within** the Keycloak container. 
HAProxy is not available there. 
You can also verify that you are accessing Keycloak directly, by checking the port you are using. If you were accessing Keycloak via HAProxy you would use port 443, not 9000 as you do now.

dzarras (approved):

PR #565, opened 2026-06-04
Periodically revoke issued credentials based on client and key storage status

### LINKED ISSUE #529: PID\Attestation issuance adaptations to TS3 v1.5

gdimtsas:
Closes #529 

In a nutshell, this PR does the following:

- Persist the client_status and the key_storage_satus along with the credential status for each issued credential
- Periodically check the client and key storage statuses of each active issued credential. If one is no longer valid, revoke the credential.
- Add statium in order to check status list tokens
- Add a new postgres adapter for persistence

An example of how credentials are stored in the database:

|id|format|type|issued_at|expires_at|notification_id|status_list_uri|status_list_index|client_status_list_uri|client_status_list_index|key_storage_status_list_uri|key_storage_status_list_index|
|--|------|----|---------|----------|---------------|---------------|-----------------|----------------------|------------------------|---------------------------|-----------------------------|
|1|dc+sd-jwt|urn:eudi:pid:1|2026-06-03 15:28:28.257 +0300|2026-07-04 15:28:28.257 +0300|0f7998f6-56be-47e3-b5dc-61aed489e685|https://issuer.eudiw.dev/token_status_list/FC/urn:eudi:pid:1/821d697b-b88b-4f5b-8f8d-5f5fa6c5b611|7489|https://issuer.eudiw.dev/token_status_list/FC/key-attestation+jwt/db47428f-a819-409d-b304-c067b1b9177f|2716|https://issuer.eudiw.dev/token_status_list/FC/key-attestation+jwt/db47428f-a819-409d-b304-c067b1b9177f|9806|


src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/ehic/IssueSdJwtVcEuropeanHealthInsuranceCard.kt
```diff
@@ -219,25 +219,39 @@ internal class IssueSdJwtVcEuropeanHealthInsuranceCard private constructor(
...
+                ehic,
+                authorizationContext.username,
+                it,
+                dateOfIssuance = dateOfIssuance,
+                dateOfExpiry = dateOfExpiry,
+            ).bind()
+            encodedCredential to null
+        }
+
+        credentialsToStatuses.forEach {
```
dzarras: @gdimtsas I'd suggest to avoid iterating over the list again. We can take advantage of the `parMap` above, and parallelize the workload a bit.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/learningcredential/IssueLearningCredential.kt
```diff
@@ -73,25 +73,33 @@ internal class IssueLearningCredential(
...
-            encodeLearningCredential(learningCredential, it, issuedAt = issuedAt, expiresAt = expiresAt).bind()
-        }.toNonEmptyListOrNull()
+        val notificationId = generateNotificationId?.invoke()
+
+        val credentialsToStatuses = holderKeys.parMap(Dispatchers.Default, 4) {
+            val encodedCredential = encodeLearningCredential(learningCredential, it, issuedAt = issuedAt, expiresAt = expiresAt).bind()
+            encodedCredential to null
+        }
+
+        credentialsToStatuses.forEach {
```
dzarras: @gdimtsas Same here, I'd suggest to avoid iterating over the list again. We can take advantage of the `parMap` above, and parallelize the workload a bit.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/mdl/IssueMobileDrivingLicence.kt
```diff
@@ -366,36 +366,40 @@ internal class IssueMobileDrivingLicence(
...
-            Unexpected("Unable to issue mDL")
+            val encodedCredential =
+                encodeMobileDrivingLicenceInCbor(licence, holderKey, issuedAt = issuedAt, expiresAt = expiresAt, statusListToken).bind()
+            encodedCredential to statusListToken
         }
 
-        val notificationId =
-            if (notificationsEnabled) generateNotificationId()
-            else null
+        credentialsToStatuses.forEach {
```
dzarras: @gdimtsas Same here, I'd suggest to avoid iterating over the list again. We can take advantage of the `parMap` above, and parallelize the workload a bit.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/persistence/JdbcIssuedCredentialRepository.kt
```diff
@@ -0,0 +1,96 @@
...
+private val log = LoggerFactory.getLogger(JdbcIssuedCredentialRepository::class.java)
+
+/**
+ * Spring Data R2DBC repository for [IssuedCredentialEntity].
+ */
+interface IssuedCredentialR2dbcRepository :
+    CoroutineCrudRepository<IssuedCredentialEntity, Long>,
+    CoroutineSortingRepository<IssuedCredentialEntity, Long> {
+
+    suspend fun findAllByNotificationId(notificationId: String): List<IssuedCredentialEntity>
```
dzarras: @gdimtsas This can be potentially fetch many entities. I'd suggest to use a Flow instead.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/pid/IssueMsoMdocPid.kt
```diff
@@ -329,40 +329,45 @@ internal class IssueMsoMdocPid(
...
+            val encodedCredential =
+                encodePidInCbor(pid, pidMetaData, holderKey, issuedAt = issuedAt, expiresAt = expiresAt, statusListToken)
+                    .also {
+                        log.info("Issued $it")
+                    }
+            encodedCredential to statusListToken
+        }
+
+        // persist issued credentials
+        credentialsToStatuses.forEach {
```
dzarras: @gdimtsas Same here, I'd suggest to avoid iterating over the list again. We can take advantage of the parMap above, and parallelize the workload a bit.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/pid/IssueSdJwtVcPid.kt
```diff
@@ -267,27 +270,29 @@ internal class IssueSdJwtVcPid(
...
-        ensureNotNull(issuedCredentials) {
-            Unexpected("Unable to issue PID")
+            encodedCredential to statusListToken
         }
 
-        val notificationId =
-            if (notificationsEnabled) generateNotificationId()
-            else null
+        // persist issued credentials
+        credentialsToStatuses.forEach {
```
dzarras: @gdimtsas Same here, I'd suggest to avoid iterating over the list again. We can take advantage of the parMap above, and parallelize the workload a bit.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/status/GetStatusListTokenWithStatium.kt
```diff
@@ -0,0 +1,56 @@
...
+ * Checks the status of a single entry in a Token Status List using the statium library.
+ */
+internal class GetStatusListTokenWithStatium(
+    private val httpClient: HttpClient,
+) : GetStatusListTokenStatus {
+
+    override suspend fun invoke(uri: URI, index: UInt): Either<Throwable, StatusListTokenStatus> =
+        Either.catch {
+            val getStatusListToken = GetStatusListToken.usingJwt(
+                clock = Clock.System,
```
dzarras: @gdimtsas I think we have a Clock bean in the Spring Context. Let's use this instead.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/status/MarkStatusAsRevokedWithExternalService.kt
```diff
@@ -0,0 +1,62 @@
...
+            .body(
+                BodyInserters.fromFormData(
+                    LinkedMultiValueMap<String, String>().apply {
+                        add(IDX_PARAM, index.toString())
+                        add(URI_PARAM, uri.toString())
+                        add(STATUS_PARAM, STATUS_REVOKED)
+                    },
+                ),
+            )
+            .awaitExchange { response ->
```
dzarras, line 47: @gdimtsas Given the return type of the function is not an Either or a Result, I'd probably use `retrieve` instead and let the `WebClient` propagate a `WebClientResponseException` instead.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/Types.kt
```diff
@@ -130,13 +128,16 @@ sealed interface CryptographicBindingMethod {
...
     val type: String,
-    val holder: String,
-    val holderPublicKeys: NonEmptyList<JWK>,
     val issuedAt: Instant,
+    val expiresAt: Instant,
     val notificationId: NotificationId? = null,
+    val statusListToken: StatusListToken?,
+    val clientStatusListToken: StatusListToken,
+    val keyStorageStatusListToken: StatusListToken,
+    val persistenceId: Long? = null,
```
dzarras: @gdimtsas Is `persistenceId` something we need in the domain?

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/RevokeCredentialsWithRevokedStatus.kt
```diff
@@ -0,0 +1,108 @@
...
+
+private val log = LoggerFactory.getLogger(RevokeCredentialsWithRevokedStatus::class.java)
+
+/**
+ * Use case that retrieves all active issued credentials and revokes any whose
+ * client status or key storage status has been revoked in the respective status list.
+ */
+class RevokeCredentialsWithRevokedStatus(
+    private val clock: Clock,
+    private val deleteExpiredIssuedCredentials: DeleteExpiredIssuedCredentials,
```
dzarras, line 40: @gdimtsas This seems to be the concern of a different job.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/RevokeCredentialsWithRevokedStatus.kt
```diff
@@ -0,0 +1,108 @@
...
+
+    suspend operator fun invoke() {
+        deleteExpiredIssuedCredentials(clock)
+        val activeCredentials = getActiveIssuedCredentials(clock)
+        log.info("Checking revocation status for {} active credential(s)", activeCredentials.size)
+
+        activeCredentials
+            .filter { it.statusListToken != null }
+            .forEach { credential ->
+                runCatching {
```
dzarras: @gdimtsas Prefer using Arrow's `result` to avoid catching `CancellationExcpetion` and other fatal exceptions that cannot be dealt with.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/RevokeCredentialsWithRevokedStatus.kt
```diff
@@ -0,0 +1,108 @@
...
+                        e.message,
+                        e,
+                    )
+                }
+            }
+    }
+
+    private suspend fun isStatusRevoked(
+        statusName: String,
+        statusListToken: Pair<URI, UInt>,
```
dzarras: @gdimtsas Why use a Pair instead of StatusListToken?

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/RevokeCredentialsWithRevokedStatus.kt
```diff
@@ -0,0 +1,108 @@
...
+            .filter { it.statusListToken != null }
+            .forEach { credential ->
+                runCatching {
+                    val clientStatusList = credential.clientStatusListToken.let {
+                        it.statusList to it.index
+                    }
+                    val keyStorageStatusList = credential.keyStorageStatusListToken.let {
+                        it.statusList to it.index
+                    }
+                    val shouldRevoke =
```
dzarras: @gdimtsas Better name this `mustRevoke`.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/RevokeCredentialsWithRevokedStatus.kt
```diff
@@ -0,0 +1,108 @@
...
+                            }
+                    }
+                }.onFailure { e ->
+                    log.error(
+                        "Unexpected error processing credential: {}",
+                        e.message,
+                        e,
+                    )
+                }
+            }
```
dzarras, line 60: @gdimtsas I'd expect a call to `getOrThrow()` somewhere.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/out/persistence/GetActiveIssuedCredentials.kt
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package eu.europa.ec.eudi.pidissuer.port.out.persistence
+
+import eu.europa.ec.eudi.pidissuer.domain.Clock
+import eu.europa.ec.eudi.pidissuer.domain.IssuedCredential
+
+fun interface GetActiveIssuedCredentials {
```
dzarras: @gdimtsas Please rename this to `GetNonExpiredIssuedCredentials`.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/out/status/MarkStatusAsRevoked.kt
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+package eu.europa.ec.eudi.pidissuer.port.out.status
+
+import arrow.core.Either
+import java.net.URI
+
+fun interface MarkStatusAsRevoked {
+
+    suspend operator fun invoke(uri: URI, index: UInt): Either<Throwable, Unit>
```
dzarras: @gdimtsas You could use `StatusListToken` as input.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerContext.kt
```diff
@@ -391,6 +399,7 @@ fun beans(clock: Clock) = BeanRegistrarDsl {
             WebClients.default(proxy = proxy)
         }
     registerBean { webClient }
+    registerBean { KtorHttpClient(JavaEngine) }
```
dzarras: @gdimtsas Please try to replicate the logic we use for instantiating the `WebClient`. When creating an `HttpClient` 
1. we must check whether the `insecure` profile is enabled configure the client accordingly 
2. we must check whether an http proxy has been configured and use it

src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerContext.kt
```diff
@@ -498,19 +507,46 @@ fun beans(clock: Clock) = BeanRegistrarDsl {
...
+        RevokeCredentialsWithRevokedStatus(
+            clock = clock,
+            deleteExpiredIssuedCredentials = bean(),
+            getActiveIssuedCredentials = bean(),
+            getStatusListTokenStatus = bean(),
+            markStatusAsRevoked = bean(),
+            deleteIssuedCredential = bean(),
+        )
+    }
+    if (enableRevocationJob) {
```
dzarras: @gdimtsas Revoking VCs is not optional task. I wouldn't allow the user to disable the job.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/input/scheduler/CredentialRevocationJob.kt
```diff
@@ -0,0 +1,43 @@
...
+ * The schedule can be configured via the `issuer.revocation.job.cron` property
+ * (default: every 8 hours).
+ */
+class CredentialRevocationJob(
+    private val revokeCredentialsWithRevokedStatus: RevokeCredentialsWithRevokedStatus,
+) {
+
+    fun run() {
+        log.info("Starting credential revocation job")
+        runBlocking {
```
dzarras: @gdimtsas I'd suggest to make this function suspending and avoid runBlocking here. Use runBlocking when you have to bridge with framework code, i.e. in `SchedulingConfigurer`

dzarras: Also, do we really need this job as a separate class?

dzarras (changes requested):
@gdimtsas Added some initial comments for your consideration.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/ehic/IssueSdJwtVcEuropeanHealthInsuranceCard.kt
gdimtsas: Done!

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/learningcredential/IssueLearningCredential.kt
gdimtsas: Done!

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/mdl/IssueMobileDrivingLicence.kt
gdimtsas: Done!

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/persistence/JdbcIssuedCredentialRepository.kt
gdimtsas: Done!

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/pid/IssueMsoMdocPid.kt
gdimtsas: Done!

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/pid/IssueSdJwtVcPid.kt
gdimtsas: Done!

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/out/persistence/GetActiveIssuedCredentials.kt
gdimtsas: Done!

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/RevokeCredentialsWithRevokedStatus.kt
gdimtsas: It is indeed. But for simplicity and to avoid race conditions I added this functionality as the first step on the same job.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/RevokeCredentialsWithRevokedStatus.kt
gdimtsas: Done!

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/RevokeCredentialsWithRevokedStatus.kt
gdimtsas: Refactored to use StatusListToken, no point in using a pair

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/out/status/MarkStatusAsRevoked.kt
gdimtsas: The idea here is to avoid importing a class from domain (StatusListToken) into the adapter, so i pass explicitly the URI and UInt

src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerContext.kt
gdimtsas: Done!

src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerContext.kt
gdimtsas: I agree. I removed the option

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/RevokeCredentialsWithRevokedStatus.kt
gdimtsas: I don't think we need one. This is a loop of all non expired credentials. Event in the case of failure for one of them, the error is logged and the processing continues with the rest. We don't want to force success for all of them.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/status/GetStatusListTokenWithStatium.kt
gdimtsas: Done!

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/input/scheduler/CredentialRevocationJob.kt
gdimtsas: @dzarras This is in fact a function that is called from `SchedulingConfigurer`, I don't think it can be made suspendable.

build.gradle.kts
```diff
@@ -88,6 +90,19 @@ dependencies {
         because("To support CNonce encryption using XC20P")
     }
 
+    implementation("org.springframework.boot:spring-boot-starter-data-r2dbc") {
+        because("Reactive database access for PostgreSQL persistence adapter")
+    }
+    implementation("org.postgresql:r2dbc-postgresql") {
```
dzarras: @gdimtsas I'd suggest we also add the R2DBC pool driver as well.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/status/MarkStatusAsRevokedWithExternalService.kt
gdimtsas: This function returns an Either

build.gradle.kts
```diff
@@ -88,6 +90,19 @@ dependencies {
...
+        because("R2DBC driver for PostgreSQL")
+    }
+    implementation("io.r2dbc:r2dbc-h2") {
```
dzarras: @gdimtsas Please mark the H2 dependencies as test dependencies.

build.gradle.kts
dzarras: Additionally, I have the impression we can mark the dependencies as runtime only. I don't think we need them in the compile time classpath.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/persistence/JdbcIssuedCredentialRepository.kt
```diff
@@ -0,0 +1,97 @@
...
+
+    @Query("DELETE FROM issued_credential WHERE expires_at <= :now")
+    suspend fun deleteAllExpiredBefore(now: OffsetDateTime): Int?
+}
+
+/**
+ * R2DBC-backed adapter implementing all issued-credential persistence ports.
+ * Compatible with PostgreSQL and H2.
+ */
+class JdbcIssuedCredentialRepository(
```
dzarras: @gdimtsas Please rename the class. It's not JDBC based. Also rename the file as well.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/Types.kt
dzarras: @gdimtsas I've been thinking about this. `IssuedCredential` is a root aggregate. It must have an entity id. 

We have two options:
1. Generate the entity id in the domain. We could create a value class that uses a UUIDv7, and use this value as a natural Id in the database alongside the autogenerated row id, or use it as the row id.
2. Use the autogenerated database id as the entity id. To use this approach we'd have to change the signature of `StoreIssuedCredential` to something like: `suspend operator fun invoke(format: Format, type: String, ..., keyStorageStatus: StatusListToken): IssuedCredential`

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/Types.kt
```diff
@@ -130,13 +128,16 @@ sealed interface CryptographicBindingMethod {
...
+data class IssuedCredential(
     val format: Format,
     val type: String,
-    val holder: String,
-    val holderPublicKeys: NonEmptyList<JWK>,
     val issuedAt: Instant,
+    val expiresAt: Instant,
     val notificationId: NotificationId? = null,
+    val statusListToken: StatusListToken?,
+    val clientStatusListToken: StatusListToken,
```
dzarras: @gdimtsas We could simplify the names to `clientStatus`, and `keyStorageStatus`. We know they are Status List Tokens from their type.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/out/persistence/GetNonExpiredIssuedCredentials.kt
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+package eu.europa.ec.eudi.pidissuer.port.out.persistence
+
+import eu.europa.ec.eudi.pidissuer.domain.Clock
+import eu.europa.ec.eudi.pidissuer.domain.IssuedCredential
+
+fun interface GetNonExpiredIssuedCredentials {
+
+    suspend operator fun invoke(clock: Clock): List<IssuedCredential>
```
dzarras: @gdimtsas I'd suggest we change the signature to `operator fun invoke(at: Instant): Flow<IssuedCredential>`. Flow because we don't know how big the list is going to be.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/Types.kt
gdimtsas: @dzarras We need a unique identifier for the stored credential, so that we can do delete them by the identifier. We don't have any other information stored so that we can identify stored credentials with. I can add a UUID that will be stored to the DB, in order to hide the table id from the domain. What do you think?

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/out/status/MarkStatusAsRevoked.kt
dzarras: Adapters can depend and use domain types.

dzarras (changes requested):

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/input/scheduler/CredentialRevocationJob.kt
dzarras: I know. My point is about where `runBlocking` should be used. `runBlocking` is used to call suspending functions from blocking fuctions, so I would expect `runBlocking` to used in `SchedulingConfigurer`.

gdimtsas: Got it, it makes sense now. I moved the runBlocking in the SchedulingConfigurer

build.gradle.kts
gdimtsas: I believe that the R2DBC pool driver is a transitive dependency of spring-boot-starter-data-r2dbc. Concerning the dependencies, they can indeed be marked as runtimeOnly.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/RevokeCredentialsWithRevokedStatus.kt
```diff
@@ -0,0 +1,102 @@
...
+    private val deleteIssuedCredential: DeleteIssuedCredential,
+) {
+
+    suspend operator fun invoke() {
+        deleteExpiredIssuedCredentials(clock)
+        val activeCredentials = getNonExpiredIssuedCredentials(clock)
+        log.info("Checking revocation status for {} active credential(s)", activeCredentials.size)
+
+        activeCredentials
+            .filter { it.statusListToken != null }
```
vafeini: Is this to filter only device bound attestations? For non device bound attestations we might still get a WIA, thus a client status that should be monitored.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/persistence/IssuedCredentialEntity.kt
```diff
@@ -0,0 +1,105 @@
...
+    @Column("status_list_uri")
+    val statusListUri: String?,
+    @Column("status_list_index")
+    val statusListIndex: Long?,
+    @Column("client_status_list_uri")
+    val clientStatusListUri: String,
+    @Column("client_status_list_index")
+    val clientStatusListIndex: Long,
+    @Column("key_storage_status_list_uri")
+    val keyStorageStatusListUri: String,
```
vafeini: Shouldn't properties `keyStorageStatusListUri` and `keyStorageStatusListIndex` be nullable for the case of non-device bound attestations?

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/RevokeCredentialsWithRevokedStatus.kt
```diff
@@ -0,0 +1,102 @@
...
+class RevokeCredentialsWithRevokedStatus(
+    private val clock: Clock,
+    private val deleteExpiredIssuedCredentials: DeleteExpiredIssuedCredentials,
+    private val getNonExpiredIssuedCredentials: GetNonExpiredIssuedCredentials,
+    private val getStatusListTokenStatus: GetStatusListTokenStatus,
+    private val markStatusAsRevoked: MarkStatusAsRevoked,
+    private val deleteIssuedCredential: DeleteIssuedCredential,
+) {
+
+    suspend operator fun invoke() {
```
vafeini, line 48: Let's consider adding some parallelism in the whole process. A proposal:

1. Parallelize all status checks for all active credentials. For each credential there are up to 2 status lookups that can happen. All lookups should be 'flattened' and executed in parallel. The accumulation of all status checks will be 

credential_id + client_status_type + lookup_result (revoked/not-revoked)

  Based on all the above it can be deducted which issued attestations should be revoked 

2. Based on the result of the above revoke attestations. These can be also parallelized.  
3. For all succeeded revocations of previous step make one batch delete in database. 

In case of failures in steps 2 and 3 we can think of applying some retry policy.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/RevokeCredentialsWithRevokedStatus.kt
dzarras: There are also short-lived attestations that don't have a Status List Token. Currently we don't have any in pid-issuer. But I'm guessing this is forward thinking.

gdimtsas: Exactly, this check is about credentials without a status list. Of course these credentials don't have to be tracked and be revoked, but they are stored in the same table just for the purpose of retrieving them during the notification request.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/persistence/IssuedCredentialEntity.kt
gdimtsas: I now made them nullable in order to be ready for non-device bound attestations

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/RevokeCredentialsWithRevokedStatus.kt
gdimtsas: I added some parallelism on the processing of the activeCredentials Flow:

```kotlin
activeCredentials
    .filter { it.status != null }
    .flatMapMerge { credential ->
        flow {
            emit(processCredential(credential))
        }
    }
    .collect { }
```

I think it is more simple to read and will complete in at least the same time as the proposal above.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/RevokeCredentialsWithRevokedStatus.kt
vafeini: Still for short-lived attestations we are receiving a WIA that has a client status. This means that there is a possibility that even though the attestation is short-lived WIA can be revoked before expiration.

gdimtsas: Your point is to remove them from the database if the WIA/KA is revoked? Because that's the only thing we can do since they dont have a status to revoke.

vafeini: Yes

gdimtsas: The checks are now done for all credentials, short lived or not. If revoked, they are deleted from the DB.

vafeini (dismissed):

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/Types.kt
```diff
@@ -127,16 +125,25 @@ sealed interface CryptographicBindingMethod {
...
     val type: String,
-    val holder: String,
-    val holderPublicKeys: NonEmptyList<JWK>,
     val issuedAt: Instant,
+    val expiresAt: Instant,
     val notificationId: NotificationId? = null,
+    val status: StatusListToken?,
+    val clientStatus: StatusListToken,
+    val keyStorageStatus: StatusListToken?,
+    val id: IssuedCredentialId? = null,
```
dzarras: @gdimtsas `id` being nullable is not OK. The identity of the root aggregate must be not-null.

* Either update `StoreIssuedCredential` to `suspend operator fun invoke(format: Format, type: String, ..., keyStorageStatus: StatusListToken): IssuedCredential` and return the `IssuedCredential` using the `id` that has been set by the persistence provider.
* Or make `IssuedCredentialId` a value class over `Uuid` and populate it using either a v4, or v7 UUID.

`StoreIssuedCredential` as it is currently defined and implemented has a major issue. It has a side-effect that is never communicated to the caller, i.e. the generation of the root aggregate id.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/out/persistence/DeleteExpiredIssuedCredentials.kt
```diff
@@ -0,0 +1,26 @@
...
+package eu.europa.ec.eudi.pidissuer.port.out.persistence
+
+import eu.europa.ec.eudi.pidissuer.domain.Clock
+
+/**
+ * Deletes all issued credentials that have expired (i.e. whose expiry is before now).
+ */
+fun interface DeleteExpiredIssuedCredentials {
+
+    suspend operator fun invoke(clock: Clock)
```
dzarras: @gdimtsas Instead of passing the `Clock`, I'd suggest we pass the `Instant` instead. It seems more straightforward to me.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/out/persistence/GetNonExpiredIssuedCredentials.kt
```diff
@@ -0,0 +1,25 @@
...
+ */
+package eu.europa.ec.eudi.pidissuer.port.out.persistence
+
+import eu.europa.ec.eudi.pidissuer.domain.Clock
+import eu.europa.ec.eudi.pidissuer.domain.IssuedCredential
+import kotlinx.coroutines.flow.Flow
+
+fun interface GetNonExpiredIssuedCredentials {
+
+    suspend operator fun invoke(clock: Clock): Flow<IssuedCredential>
```
dzarras: @gdimtsas Same here. I'd prefer if we passed an `Instant` instead.

dzarras (changes requested):
@gdimtsas Feel free to ignore the comments about `Clock`/`Instant`. My main concern is the `Id` of the `IssuedCredential`. 

@vafeini any comments/suggestions about it?

dzarras (dismissed):

dzarras (approved):

PR #568, opened 2026-06-12
Add Status List service to docker compose

dzarras:
This PR add the Status List service to the docker compose file used for local development.

Unfortunately the service does include an endpoint to fetch the a Status List either in JWT or CWT format, so as a workaround we add a new endpoint when building the docker image.
Additionally, a workaround has been applied to fix the hardcoded URL in the OpenAPI specification of the service.


docker-compose/status-list/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM python:3.14.6-alpine3.24
+ARG VERSION
+
+RUN \
+    wget https://github.com/eu-digital-identity-wallet/eudi-srv-statuslist-py/archive/refs/tags/$VERSION.zip -O /tmp/app.zip && \
+    unzip /tmp/app.zip && \
+    mv eudi-srv-statuslist-py-* status-list && \
+    rm /tmp/app.zip
```
atsant, line 8: @dzarras I would suggest to split this to an `ADD` and then the `RUN` for the different operations.

atsant (changes requested):
Minor comment just to make the DockerFile a little bit more understandable.

atsant (approved):

PR #586, opened 2026-06-25
Refactor format encoders and Issuers

babisRoutis:
The PR makes the following improvements:

## Unification of format encoders

```kotlin
data class AttestationAttributes<out Data>(
    val attributes: Data,
    val issuedAt: Instant,
    val expiresAt: Instant,
    val notBefore: Instant? = null,
    val deviceKey: JWK? = null,
    val status: StatusListToken? = null,
    val jwtId: String? = null,
)

fun interface EncodeAttestationAttributes<in Attr> {
    suspend operator fun invoke(attestationAttributes: AttestationAttributes<Attr>): JsonElement
}
```

## Unification of MDoc Issuers

All mdoc issuers (PID and mDL) share a common implementation `IssueMdoc`.

Minor improvements
- Partially cleanup the huge configuration file of the application


src/main/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/format/EncodeAttestationAttributes.kt
```diff
@@ -0,0 +1,35 @@
...
+import kotlin.time.Instant
+
+data class AttestationAttributes<out Data>(
+    val attributes: Data,
+    val issuedAt: Instant,
+    val expiresAt: Instant,
+    val notBefore: Instant? = null,
+    val deviceKey: JWK? = null,
+    val status: StatusListToken? = null,
+    val jwtId: String? = null,
```
dzarras: Since `jwtId` is specific to JWT-based attestations, wouldn't it make sense to make it part of `attributes` instead?

src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerContext.kt
```diff
@@ -875,31 +565,32 @@ fun beans(
...
+        val dPoPConfigurationProperties =
+            if (AccessTokenType.DPoP == accessTokenType || AccessTokenType.BearerAndDPoPIfAvailable == accessTokenType) {
+                val algorithms =
+                    runBlocking {
+                        val authorizationServerMetadata =
+                            URI.create(env.getRequiredProperty("issuer.authorizationServer.metadata"))
+                        webClient.authorizationServerSupportedDPoPJWSAlgorithms(authorizationServerMetadata)
+                    }
 
-            registerBean { DPoPConfigurationProperties(algs, realm) }
```
dzarras, line 894: As it currently is `DPoPConfigurationProperties` must be registered as a bean when DPoP must be enabled. If it's not registered as a bean, it won't be picked up at #680, and then subsquently at #688.

dzarras (changes requested):

src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerContext.kt
dzarras: We should propbably make it a val similar to `enableBearerTokenAuthentication`, and use it directly, instead of loading it as a bean.

dzarras (approved):
Verified with OpenId4VCI library examples.

PR #589, opened 2026-06-25
Remove support for Bearer authorization

### LINKED ISSUE #587: Remove support for Bearer Authentication
Currently the issuer can be configured to support :
- Bearer authentication (Using `access_token` of type `bearer`), or
- DPoP authentication  (where the type of the `access_token` is `DPoP` ), or
- Both

According to 
- HAIP v1 `DPoP` is required 
- TS3 requires DPoP for Attestation-Based Client Authentications, which means that issued `access_token` are of type `DPoP`
- ETSI 119 472-3 requires DPoP.

Thus, the credential issuer MUST support only `DPoP`  access tokens.

babisRoutis:
Closes #587

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerContext.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerContext.kt
@@ -627,10 +627,10 @@ fun beans(
             }
 
             val introspectionProperties = bean<OAuth2ResourceServerProperties>()
-            val introspectionEndpoint = introspectionProperties.opaquetoken.introspectionUri
-            checkNotNull(introspectionEndpoint) {
-                "missing spring.security.oauth2.resourceserver.opaquetoken.introspection-uri configuration property"
-            }
+            val introspectionEndpoint =
+                checkNotNull(introspectionProperties.opaquetoken.introspectionUri) {
+                    "missing spring.security.oauth2.resourceserver.opaquetoken.introspection-uri configuration property"
+                }
             val introspector =
                 SpringReactiveOpaqueTokenIntrospector(
                     introspectionEndpoint,
@@ -644,11 +644,6 @@ fun beans(
                         }.build(),
                 )
 
-            val entryPoints = mutableListOf<DelegatingServerAuthenticationEntryPoint.DelegateEntry>()
-            val accessDeniedHandlers =
-                mutableListOf<ServerWebExchangeDelegatingServerAccessDeniedHandler.DelegateEntry>()
-            val filters = mutableListOf<Pair<SecurityWebFiltersOrder, WebFilter>>()
-
             log.info("Enabling DPoP AccessToken support")
 
             val enableDPoPNonce = env.getProperty<Boolean>("issuer.dpop.nonce.enabled") ?: true
@@ -660,23 +655,8 @@ fun beans(
                     DPoPNoncePolicy.Disabled
                 }
 
-            val entryPoint =
-                DPoPTokenServerAuthenticationEntryPoint(dPoPConfigurationProperties.realm, dpopNonce, bean())
+            val entryPoint = DPoPTokenServerAuthenticationEntryPoint(dPoPConfigurationProperties.realm, dpopNonce, bean())
             val tokenConverter = ServerDPoPAuthenticationTokenAuthenticationConverter()
-
-            entryPoints.add(
-                DelegatingServerAuthenticationEntryPoint.DelegateEntry(
-                    AuthenticationConverterServerWebExchangeMatcher(tokenConverter),
-                    entryPoint,
-                ),
-            )
-
-            accessDeniedHandlers.add(
-                ServerWebExchangeDelegatingServerAccessDeniedHandler.DelegateEntry(
-                    AuthenticationConverterServerWebExchangeMatcher(tokenConverter),
-                    DPoPTokenServerAccessDeniedHandler(dPoPConfigurationProperties.realm),
-                ),
-            )
 
             val dpopFilter =
                 run {
@@ -699,7 +679,7 @@ fun beans(
                         setAuthenticationFailureHandler(ServerAuthenticationEntryPointFailureHandler(entryPoint))
                     }
                 }
-            filters.add(SecurityWebFiltersOrder.AUTHENTICATION to dpopFilter)
+            http.addFilterAfter(dpopFilter, SecurityWebFiltersOrder.AUTHENTICATION)
 
             if (dpopNonce is DPoPNoncePolicy.Enforcing) {
                 val dpopNonceFilter =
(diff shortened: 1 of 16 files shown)


src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/GetProtectedResourceMetadata.kt
```diff
@@ -54,26 +53,19 @@ data class ProtectedResourceMetadataTO(
...
-            dpopSigningAlgorithmsSupported = dPoPSigningAlgorithmsSupported,
-            dpopBoundAccessTokenRequired = dPoPBoundAccessTokenRequired,
-        )
-    }
+    fun unsigned(): ProtectedResourceMetadataTO =
+        with(credentialIssuerMetadata) {
+            val authorizationServers = authorizationServers.map { it.externalForm }.toNonEmptyListOrNull()
+            val scopes = attestationIssuers.map { it.configuration.scope.value }.distinct().toNonEmptyListOrNull()
+            val dpopAlgs = dPoPConfigurationProperties.algorithms.map { it.name }
+            ProtectedResourceMetadataTO(
```
dzarras, line 49: We could also remove the `bearerMethodsSupported` property from `ProtectedResourceMetadataTO` and `BearerMethodTO` to avoid setting it by accident.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/GetCredentialIssuerMetaData.kt
```diff
@@ -26,7 +26,7 @@ import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.json.*
 
-class GetCredentialIssuerMetaData(
+class GetCredentialIssuerMetaData constructor(
```
dzarras, line 29: Did you mean to add a visibility modifier here as well?

src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerEnvironmentExtensions.kt
```diff
@@ -317,7 +310,7 @@ internal suspend fun WebClient.authorizationServerSupportedDPoPJWSAlgorithms(aut
         }.getOrElse {
```
dzarras: Given we require the authorization server to support DPoP, wouldn't it make sense to fail if it's not supported? 

Why let the issuer complete startup when no-one will be able to access it? 

I'd probably replace `Either.catch` simply with
```kotlin
val metadata =
                get()
                    .uri(authorizationServerMetadata)
                    .accept(MediaType.APPLICATION_JSON)
                    .retrieve()
                    .bodyToMono<String>()
                    .timeout(5.seconds.toJavaDuration())
                    .awaitSingle()
            checkNotNull(OIDCProviderMetadata.parse(metadata).dPoPJWSAlgs).toNonEmptySetOrThrow()
```

src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerContext.kt
```diff
@@ -664,105 +649,71 @@ fun beans(
...
-                            DPoPProtectedResourceRequestVerifier(
-                                dPoPConfigurationProperties.algorithms,
-                                15.seconds.inWholeSeconds,
-                                30.seconds.inWholeSeconds,
-                                InMemoryDPoPSingleUseChecker(
-                                    60.seconds.inWholeSeconds,
-                                    10.minutes.inWholeSeconds,
-                                ),
-                            )
+            entryPoints.add(
```
dzarras: `entryPoints` no longer needs to be a mutable list. We could initialize it here as `val entryPoints = listOf(...)`

src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerContext.kt
```diff
@@ -664,105 +649,71 @@ fun beans(
...
+                DelegatingServerAuthenticationEntryPoint.DelegateEntry(
+                    AuthenticationConverterServerWebExchangeMatcher(tokenConverter),
+                    entryPoint,
+                ),
+            )
 
-                        val authenticationManager =
-                            DPoPTokenReactiveAuthenticationManager(introspector, dPoPVerifier, dpopNonce, bean())
+            accessDeniedHandlers.add(
```
dzarras: Same here. `val accessDeniedHandlers = listOf(...)`

src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerContext.kt
```diff
@@ -664,105 +649,71 @@ fun beans(
                 mutableListOf<ServerWebExchangeDelegatingServerAccessDeniedHandler.DelegateEntry>()
             val filters = mutableListOf<Pair<SecurityWebFiltersOrder, WebFilter>>()
```
dzarras: `filters` could be removed.

What we now need is:
```kotlin

http.addFilterAfter(dpopFilter, SecurityWebFiltersOrder.AUTHENTICATION)
if (null != dpopNonceFilter) {
    http.addFilterAt(dpopNonceFilter, SecurityWebFiltersOrder.LAST)
}
```

src/test/kotlin/eu/europa/ec/eudi/pidissuer/adapter/out/attestation/mdl/GetMobileDrivingLicenceDataMockTest.kt
```diff
@@ -54,7 +54,7 @@ internal class GetMobileDrivingLicenceDataMockTest {
             val authorizationContext =
                 AuthorizationContext(
```
dzarras, line 55: We should also limit `accessToken` in `AuthorizationContext` to `DPoPAccessToken` as well

dzarras (changes requested):

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/GetCredentialIssuerMetaData.kt
babisRoutis: No. I was looking for the use of the constructor :smile:

src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerEnvironmentExtensions.kt
babisRoutis: In general, I think we should remove this lookup during application bootstrap.

This should be an out of band, common knowledge, between two components that are co-located.

dzarras (approved):

PR #592, opened 2026-06-26
Reduce configuration complexity

babisRoutis:
The PR tries to reduce the complexity of `IssuerContext` by extracting security configuration to a new file.

### DIFF
--- a/src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerContext.kt
+++ b/src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerContext.kt
@@ -68,13 +68,13 @@ fun beans(
     val credentialsOfferUri = env.getRequiredProperty("issuer.credentialOffer.uri")
     val issuerKeystore: KeyStore by lazy { keystore(env) }
     val getIssuerSigningKey = loadOrGenerateIssuerSigningKey(clock, env, issuerPublicUrl) { issuerKeystore }
-    val proxy = env.httpProxy()
 
     registerBean { env.dPoPConfigurationProperties() }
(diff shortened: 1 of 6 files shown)


src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerContext.kt
```diff
@@ -735,3 +486,34 @@ fun beans(
         }
     }
 }
+
+private fun loadNonceEncryptionKey(
```
dzarras: This method could also be defined as an extension of `Environment` and have a lambda parameter that returns the `KeyStore` for uniformity.
It could also be moved in `IssuerEnvironmentExtensions.kt` as well.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerContext.kt
```diff
@@ -735,3 +486,34 @@ fun beans(
...
+                log.info("Loading Nonce encryption key from keystore")
+                val nonceEncryptionKey = issuerKeystore().loadJwk(env, "issuer.nonce.encryption-key")
+                require(nonceEncryptionKey is ECKey) { "Only ECKey are supported for encryption" }
+                nonceEncryptionKey
+            }
+        }
+    return NonceEncryptionKey(encryptionKey)
+}
+
+private fun Environment.batchCredentialIssuance(): BatchCredentialIssuance {
```
dzarras: This function could be move to `IssuerEnvironmentExtensions.kt` as well.

dzarras (dismissed):
Two minor comments for your consideration. Other than that, looks good to me.

PR #598, opened 2026-07-01
Security updates

dzarras:
This PR:

1. Refactors Spring Security configuration. We now configure 2 distinct WebSecurityFilterChains, one for UI paths, one for API paths.
2. Enables CORS, and CSRF for UI paths.
3. Updates Thymeleaf templates, and removes links to external resources. -*This breaks the preview, when not running the web server.*-
4. Adds a Content Security Policy to UI paths only.
5. Adds test cases to verify the UI paths contain the CSP header, while ensuring it is not present in API paths.
6. Adds a configurable allow-list for schemes that can be used with custom Credential Offer URIs.
7. Adds test cases to verify a Credential Offer is generated only when using an allowed scheme.


src/test/kotlin/eu/europa/ec/eudi/pidissuer/port/input/CreateCredentialsOfferTest.kt
```diff
@@ -0,0 +1,101 @@
...
+    @Test
+    fun `credential offer must be created when using an allowed credential offer uri scheme`() =
+        runTest {
+            val schemes =
+                nonEmptySetOf(
+                    "https",
+                    "openid-credential-offer",
+                    "haip-vci",
+                    "eu-eaa-offer",
+                ).flatMap { nonEmptySetOf(it, it.uppercase()) }
```
babisRoutis, line 56: Why uppercase?

README.md
```diff
@@ -491,6 +491,10 @@ Variable: `ISSUER_CREDENTIALOFFER_URI`
 Description: URI to use when generating Credential Offers.    
 Default value: `openid-credential-offer://`
 
+Variable: `ISSUER_CREDENTIALOFFER_ALLOWEDSCHEMES`    
+Description: Comma-separated list of allowed schemes for using custom URIs for Credential Offers.       
+Default value: `https,openid-credential-offer,haip-vci,eu-eaa-offer`
```
babisRoutis: Please consider using a default value `eu-eaa-offer` and `https`

src/main/resources/application.properties
```diff
@@ -55,6 +55,7 @@ issuer.learningCredential.validity=P31D
 issuer.learningCredential.notifications.enabled=true
 issuer.learningCredential.jwtProofs.supportedSigningAlgorithms=ES256
 issuer.credentialOffer.uri=openid-credential-offer://
+issuer.credentialOffer.allowedSchemes=https,openid-credential-offer,haip-vci,eu-eaa-offer
```
babisRoutis: Limit the defaults to `eu-eaa-offer` and `https` (for web wallets)

src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerContext.kt
```diff
@@ -289,7 +294,8 @@ fun beans(
         GetDeferredCredential(bean(), bean(), bean())
     }
     registerBean {
-        CreateCredentialsOffer(bean(), credentialsOfferUri)
+        val allowedSchemes = env.getRequiredProperty<Set<String>>("issuer.credentialOffer.allowedSchemes").toNonEmptySetOrThrow()
```
babisRoutis: That's not safe in my opinion.

The configuration option has to be validated by an "good", closed set of schemes.
Consider, modeling the supported scheme as an enumeration and the allowedSchemes would be a non empty set of enumerated values

In this way, the configuration cannot receive get arbitrary values

src/test/kotlin/eu/europa/ec/eudi/pidissuer/port/input/CreateCredentialsOfferTest.kt
dzarras: The scheme of a URI is case-insenstive. Thus the extra test values.

README.md
dzarras: done

src/main/resources/application.properties
dzarras: done

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/CreateCredentialsOffer.kt
```diff
@@ -36,10 +36,11 @@ import java.net.URI
 class CreateCredentialsOffer(
     private val metadata: CredentialIssuerMetaData,
     private val credentialsOfferUri: String,
```
babisRoutis: This needs a better name and a more appropriate type.
Consider
- change the name to `defaultCredentialOfferUri` of type `Uri` or (URI)
- Add an init { } check that the `defaultCredentialOfferUri` is indeed a URI that has a scheme within the SupportedCredentialOfferScheme enum class

src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerContext.kt
dzarras: We could introduce an enumeration with the supported Credential Offer URI schemes, so that it's not completely open ended. 
This way we limit though wallets that use other custom uri schemes, but I think it's an acceptable compromise.

What do you think?

src/main/kotlin/eu/europa/ec/eudi/pidissuer/port/input/CreateCredentialsOffer.kt
dzarras: done

babisRoutis (approved):

PR #608, opened 2026-07-02
Improve IntelliJ support when using Spring Boot with BeanDefinitionDsl

jtonic:


src/test/kotlin/eu/europa/ec/eudi/pidissuer/TestSupport.kt
```diff
@@ -13,20 +13,17 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-@file:Suppress("ktlint:standard:max-line-length")
+@file:Suppress("ktlint:standard:max-line-length", "ktlint:standard:filename")
```
dzarras: Given this file now contains only `PidIssuerApplicationTest`, please rename the file accordingly instead of suppressing the ktlint error.

src/test/kotlin/eu/europa/ec/eudi/pidissuer/adapter/input/web/WalletApiTest.kt
```diff
@@ -77,8 +79,7 @@ import kotlin.time.Instant
  */
 @Suppress("SpringJavaInjectionPointsAutowiringInspection", "ProtectedInFinal")
 @PidIssuerApplicationTest(
-    classes = [BaseWalletApiTest.WalletApiTestConfig::class],
-    initializers = [TestMocksInitializer::class],
+    classes = [BaseWalletApiTest.WalletApiTestConfig::class, AppBeans::class],
```
dzarras: Please remove `AppBeans` from `classes`.

Annotate `PIdIssuerApplicationTest` with `@Import(AppBeans::class)` instead, so that all integration tests that make use of `PidIssuerApplicationTest` initialize the beans defined in `AppBeans`.

dzarras: Come to think of it, `PidIssuerApplicationTest` uses `PidIssuerApplication` as its configuration class. Is including `AppBeans` necessary?

src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerContext.kt
```diff
@@ -64,307 +64,306 @@ import kotlin.time.toKotlinDuration
...
-fun beans(
-    clock: Clock,
-    timeZone: TimeZone,
-) = BeanRegistrarDsl {
-    val issuerPublicUrl = env.readRequiredUrl("issuer.publicUrl", removeTrailingSlash = true)
-    val issuerKeystore: KeyStore by lazy { keystore(env) }
-    val getIssuerSigningKey = loadOrGenerateIssuerSigningKey(clock, env, issuerPublicUrl) { issuerKeystore }
+internal class AppBeans :
+    BeanRegistrarDsl({
+        val clock: Clock.System = Clock.System
```
dzarras, line 69: No need to define a type for this variable.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerContext.kt
```diff
@@ -64,307 +64,306 @@ import kotlin.time.toKotlinDuration
...
-    //
-    with(InMemoryDeferredCredentialRepository(mutableMapOf(), clock)) {
-        registerBean { GenerateTransactionId.Random }
-        registerBean { storeDeferredCredential }
-        registerBean { loadDeferredCredentialByTransactionId }
-    }
+        //
+        // Deferred Credentials
+        //
+        with(InMemoryDeferredCredentialRepository(mutableMapOf(), clock)) {
```
dzarras: Could you please update how this bean is initialized, in order to use the `bean` of `Clock` instead?

I suggest a pattern similar to how `R2dbcIssuedCredentialRepository` is initialized.

dzarras (changes requested):

src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerContext.kt
jtonic: Unfortunately this is necessary. 
There are some statements in the AppBeans constructor that are not registerBeanS.
So, what I've done:
1. I put bean() wherever clock was part of registerBean,
2. I let clock where it is part of regular methods or `with` scope function.

src/test/kotlin/eu/europa/ec/eudi/pidissuer/TestSupport.kt
jtonic: This was my intention, but considering I wanted to kept the conflicts with 0.10.0 at minimum I've retained myself in renaming this kt file.
I will do it!
Thanks

src/test/kotlin/eu/europa/ec/eudi/pidissuer/adapter/input/web/WalletApiTest.kt
jtonic: Good catch.

Thanks

src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerContext.kt
jtonic: > Could you please update how this bean is initialized, in order to use the `bean` of `Clock` instead?
> 
> I suggest a pattern similar to how `R2dbcIssuedCredentialRepository` is initialized.

Done! Great suggestion.
But still remains the last reference to the `clock` local variable, which is a function argument.

```kotlin
        val getIssuerSigningKey = loadOrGenerateIssuerSigningKey(clock, env, issuerPublicUrl) { issuerKeystore }
```

Any suggestion about this, to get rid of the clock local variable?

src/test/kotlin/eu/europa/ec/eudi/pidissuer/TestSupport.kt
jtonic: Fixed!

src/test/kotlin/eu/europa/ec/eudi/pidissuer/adapter/input/web/WalletApiTest.kt
jtonic: Fixed!

src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerContext.kt
jtonic: Fixed but still waiting for a reply in regard to the clock local variable removal.

dzarras: Leave the local variable as is for the moment.

We are in the process of dropping support for generating random signing keys and self-signed certificates.
Thus this function will no longer require a clock parameter.

For reference check this: https://github.com/eu-digital-identity-wallet/eudi-srv-pid-issuer/issues/610

dzarras (approved):

dzarras:
Again, 

Thanks a lot for your support and contributions @jtonic

PR #613, opened 2026-07-03
Drop support for generating random keys

### LINKED ISSUE #610: Drop support for generating random signing keys and Access certificates
Currently, pid-issuer allows either:
* generating a random signing key and self-signed certificate
* loading a signing key and certificate from a keystore

Same logic applies to Access certificate as well.

To reduce complexity, moving forward, we should remove support for generating a random signing key and self-signed certificate.

For local development, we already have a keystore in place.

### LINKED ISSUE #612: Drop support for generating random Credential Request encryption keys and nonce encryption key
Currently, pid-issuer for Credential Request encryption key allows:
* either generating a random key 
* or loading an encryption key from a keystore

Given pid-issuer might be deployed on a clustered environment, where all instances must advertise the same Credential Request encryption key, we should remove support for generating random keys.

This will also reduce complexity in the codebase.

For local development we already have a Credential Request encryption key in the keystore.

Generating Nonce encryption key should be dropped as well.

atsant:
Properties affected:
- Credential Request Encryption keys
- Access Certificate
- Signing Keys
- Nonce encryption key

Now all of the above are loaded exclusively from keystore.

Closes #612 #610


README.md
```diff
@@ -629,9 +594,6 @@ Example: `passwd`
 When either PID issuance in SD-JWT is enabled, or the internal MSO MDoc encoder is used, an EC Key is required
```
dzarras: We have to rephrase this section a bit:

`EC Keys are required to sign the issued credentials`

README.md
```diff
@@ -629,9 +594,6 @@ Example: `passwd`
 When either PID issuance in SD-JWT is enabled, or the internal MSO MDoc encoder is used, an EC Key is required 
 for signing the issued credentials.
 
-By default, the server generates a random EC Key alongside a self-signed certificate using the *P-256/secp256r1* 
-curve on startup. If the server is restarted, a new EC Key and self-signed certificate is generated.
-
 > [!TIP]
 > In case you opt to use your own EC Key and certificate, 
 > make sure to use an EC Key that uses one of the following curves:
```
dzarras: This as well:

`Make sure to use EC Keys that uses one of the following curves:`

dzarras (changes requested):

dzarras (approved):

PR #623, opened 2026-07-09
Introduce `issuer_info` to the credential issuer metadata carrying the WRPRC

### LINKED ISSUE #622: Introduce WRPRC to the attestation provider (`issuer_info`)
We need to introduce the `issuer_info` on pid issuer as defined in `ETSI TS 119 472-3` and profiled in `CIR 2024/2982`.

### Difference between `verifier_info` and `issuer_info`

The model for `issuer_info` is the same as per `verifier_info`  as menitoned in [OpenID4VP spec](https://openid.net/specs/openid-4-verifiable-presentations-1_0-final.html#section-5.1-2.10.1) with the following alterations:
- Does not include `credential_ids` field.
-  One of the elements in the issuer_info array parameter shall contain the PID/EAA Provider's registration certificate.
-  The value of the format member of the element in the `issuer_info` array 
parameter that contains the PID/EAA Provider's registration certificate shall be "registration_cert". 
-  The value of the data member of the element in the issuer_info array 
parameter that contains the PID/EAA Provider's registration certificate shall be the registration certificate of the 
PID/EAA Provider. 

### Adverising `issuer_info`

`issuer_info` should be advertised in issuer's metadata.

### Example format

```json
{
  "issuer_info": [
    {
      "format": "registration_cert",
      "data": "eyJhbGciOiJFUzI1...EF0RBtvPClL71TWHlIQ"
    }
  ]
}
```

atsant:
Closes #622


atsant:
~Currently, only one `registration_certificate` can be loaded but multiple may be needed. Going back to draft in order to be able to load multiple JWT as `registration_certificate` and separate `issuer_info` entries.~

babisRoutis:
> Currently, only one `registration_certificate` can be loaded but multiple may be needed. Going back to draft in order to be able to load multiple JWT as `registration_certificate` and separate `issuer_info` entries.

I cannot find a reason for multiple WRPRC in this attestation provider case.

Can you please reconsider this?
Let's have a single WRPRC for the issuer

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/CredentialIssuerMetaData.kt
```diff
@@ -301,3 +302,8 @@ value class PreferredKeyStorageStatusPeriod(
         require(value >= 31.days) { "Preferred key storage status period must be at least 31 days" }
     }
 }
+
+data class IssuerInfo(
```
babisRoutis: Some comments for your consideration:
- The name is not correct. `IssuerInfo` is the array, not the elements of it
- Given that this class is part of the domain, you can rename it to `WRPRC` or `AttestationProviderRegistrationCertificate` or `EAAProviderRegCert`
- The `format` attribute is a serialization concern. Perhaps, there is no need to have it here.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/CredentialIssuerMetaData.kt
```diff
@@ -240,6 +240,7 @@ data class CredentialIssuerMetaData(
     val display: List<CredentialIssuerDisplay> = emptyList(),
     val attestationIssuers: NonEmptyList<AttestationIssuer>,
     val preferredClientStatusPeriod: PreferredClientStatusPeriod,
+    val issuerInfo: NonEmptyList<IssuerInfo>,
```
babisRoutis: Consider replacing this with an attribute named `registrationCertificate` of type `EAAProviderRegCert`
(Check my comment bellow, about the later)

src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerConfigurationProperties.kt
```diff
@@ -54,6 +65,50 @@ internal data class IssuerMetadataProperties(
...
 }
 
+@ConfigurationProperties("issuer.issuerinfo")
+internal data class UserInfoProperties(
+    val registrationCertificate: List<String>,
+) {
+    init {
+        registrationCertificate
+            .forEach { registrationCertificate ->
+                val jwt = SignedJWT.parse(registrationCertificate)
```
babisRoutis: This code is not configuration related. 

Introduce an interface like `EAAProviderRegistrationCertificateValidator` and move there any checks of the WRPRC.


Such an interface could have the following method

`suspend verify(accessCertificate: X509Certificate, registrationCertificate: String): Boolean`

or 

`suspend verify(accessCertificate: X509Certificate, registrationCertificate: SignedJWT): Boolean`

README.md
```diff
@@ -549,6 +549,10 @@ Variable: `ISSUER_NONCE_ENCRYPTION_KEY_PASSWORD`
 Description: Password of the EC key-pair for nonce encryption.  
 Default value: N/A  
 
+Variable: `ISSUER_VERIFIERINFO_RRREGISTRATIONCERTIFICATE_XX`
```
babisRoutis: Just keep one

src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerConfigurationProperties.kt
atsant: @babisRoutis 
While the interface would be a cleaner approach, we wouldn't be able to restrict the initialization of the domain object.

Instead, I went with a private constructor and a factory method in the Wrprc value class. This prevents it from being initialized with random values, allowing us to enforce the required restrictions and protect the domain object.

dzarras:
@atsant Could you please target and rebase against `release/0.11.0`?

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/CredentialIssuerMetaData.kt
```diff
@@ -301,3 +309,38 @@ value class PreferredKeyStorageStatusPeriod(
...
+value class Wrprc private constructor(
+    val certificate: SignedJWT,
+) {
+    companion object {
+        fun of(registrationCertificate: String): Wrprc {
+            val jwt = SignedJWT.parse(registrationCertificate)
+            val x5c = jwt.header.x509CertChain
+            require(!x5c.isNullOrEmpty()) { "WRPRC must contain a valid certificate chain" }
+
+            val chain = X509CertChainUtils.parse(x5c)
```
dzarras: @atsant Could you please verify what `parse` does in case of an error? I recall it returns null entries in some cases.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/CredentialIssuerMetaData.kt
```diff
@@ -301,3 +309,38 @@ value class PreferredKeyStorageStatusPeriod(
...
+            val leafCert = chain.first()
+
+            val verifier: JWSVerifier =
+                when (val publicKey = leafCert.publicKey) {
+                    is ECPublicKey -> {
+                        val curve =
+                            Curve.forECParameterSpec(publicKey.params)
+                                ?: error("Unsupported EC curve for WRPRC leaf certificate")
+                        ECDSAVerifier(ECKey.Builder(curve, publicKey).build())
```
dzarras: @atsant Isn't there a constructor of the `ECDSAVerifier` that accepts directly the `ECPublicKey`? If I'm not mistaken converting to `ECKey` is not necessary.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/CredentialIssuerMetaData.kt
```diff
@@ -301,3 +309,38 @@ value class PreferredKeyStorageStatusPeriod(
...
+        fun of(registrationCertificate: String): Wrprc {
+            val jwt = SignedJWT.parse(registrationCertificate)
+            val x5c = jwt.header.x509CertChain
+            require(!x5c.isNullOrEmpty()) { "WRPRC must contain a valid certificate chain" }
+
+            val chain = X509CertChainUtils.parse(x5c)
+            val leafCert = chain.first()
+
+            val verifier: JWSVerifier =
+                when (val publicKey = leafCert.publicKey) {
```
dzarras: @atsant I'd suggest removing the `when`. It can be replaced with a single requirement. i.e. `require(publicKey is ECPublicKey)`.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/CredentialIssuerMetaData.kt
```diff
@@ -301,3 +309,38 @@ value class PreferredKeyStorageStatusPeriod(
         require(value >= 31.days) { "Preferred key storage status period must be at least 31 days" }
     }
 }
+
+@JvmInline
+value class Wrprc private constructor(
+    val certificate: SignedJWT,
+) {
+    companion object {
+        fun of(registrationCertificate: String): Wrprc {
```
dzarras: @atsant Given `of` can throw at multiple points, I'd suggest the following pattern instead.

Rename `of`() to `tryParse()` and return `Result<Wrprc>` instead.
Then you can introduce `of(): Wrprc` which calls `tryParse().getOrThrow()`, and `ofOrNull(): Wrprc?` which calls `tryParse().getOrNull()`

You provide more flexibility to the caller this way.

docker-compose/docker-compose.yaml
```diff
@@ -140,6 +140,7 @@ services:
       - ISSUER_METADATA_DISPLAY_0_LOCALE=en
       - ISSUER_METADATA_DISPLAY_0_LOGO_URI=https://eudiw.dev/ic-logo.svg
       - ISSUER_METADATA_DISPLAY_0_LOGO_ALTERNATIVE_TEXT=EU Digital Identity Wallet Logo
+      - ISSUER_WRPRC=
```
dzarras: @atsant This configuration will never work though. The context will never initialize. I'd suggest adding an example SignedJWT that satisfies the Wrprc invariants.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerEnvironmentExtensions.kt
```diff
@@ -21,8 +21,12 @@ import com.nimbusds.jose.CompressionAlgorithm
 import com.nimbusds.jose.EncryptionMethod
 import com.nimbusds.jose.JWEAlgorithm
 import com.nimbusds.jose.JWSAlgorithm
+import com.nimbusds.jose.JWSVerifier
+import com.nimbusds.jose.crypto.ECDSAVerifier
```
dzarras: @atsant Could you please optimize the imports? I'm seeing new imports, but no new usages.

src/test/kotlin/eu/europa/ec/eudi/pidissuer/port/input/GetDeferredCredentialTest.kt
```diff
@@ -77,6 +78,9 @@ class GetDeferredCredentialTest {
             ): EncryptedJWT = throw UnsupportedOperationException("Not expected in this test")
         }
 
+    private val jwtResourcePath = "/eu/europa/ec/eudi/pidissuer/adapter/out/jose/x5c/registration-certificate.jwt"
```
dzarras: @atsant No need for this variable to be available everywhere. I'd suggest removing it and pass the value directly in loadResource below. Besides that, the variable name is way too generic. It can easily confuse someone.

src/test/kotlin/eu/europa/ec/eudi/pidissuer/port/input/IssueCredentialTest.kt
```diff
@@ -64,6 +65,9 @@ class IssueCredentialTest {
                 CredentialResponse.Issued(nonEmptyListOf(JsonPrimitive("test-credential")))
         }
 
+    private val jwtResourcePath = "/eu/europa/ec/eudi/pidissuer/adapter/out/jose/x5c/registration-certificate.jwt"
```
dzarras: @atsant Same here. Please remove this variable.

README.md
```diff
@@ -549,6 +549,10 @@ Variable: `ISSUER_NONCE_ENCRYPTION_KEY_PASSWORD`
 Description: Password of the EC key-pair for nonce encryption.  
 Default value: N/A  
 
+Variable: `ISSUER_WRPRC`  
+Description: The JWT formatted registration certificate of Credential Issuer.
```
dzarras: @atsant Please try to be more precise in your descriptions, and use terms that are already used in specs and are well-known. For instance `JWT formatted` does not mean anything.

A much better description would be: `The Registration Certificate of the Credential Issuer serialized using JWS Compact Serialization`

src/main/kotlin/eu/europa/ec/eudi/pidissuer/IssuerContext.kt
```diff
@@ -247,6 +247,9 @@ internal class AppBeans :
             val preferredClientStatusPeriod =
                 bean<IssuerMetadataProperties>().preferredClientStatusPeriod.toKotlinDuration()
 
+            val registrationCertificateProperty = env.getRequiredProperty("issuer.wrprc")
```
dzarras: @atsant I'd suggest removing this variable as well.

dzarras (changes requested):

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/CredentialIssuerMetaData.kt
```diff
@@ -301,3 +309,36 @@ value class PreferredKeyStorageStatusPeriod(
         require(value >= 31.days) { "Preferred key storage status period must be at least 31 days" }
     }
 }
+
+@JvmInline
+value class Wrprc private constructor(
+    val certificate: SignedJWT,
+) {
+    companion object {
+        fun tryParse(registrationCertificate: String): Result<Wrprc> =
```
babisRoutis: Please consider the following

1) Why `tryParse` is public? It seems just an implementation detail
2) Why use `result{ }`? 
3) Is there a use for `orNull`?

Especially for the (2) it is known that Result<T> is problematic, catching exceptions that shouldn't be caught, being unrecoverable.

Furthermore, the two uses of `tryParse` outcome, immediately throw the exception, without providing a recovery logic (because there isn't)


In general, if `orNull` is not being used, then let `tryParse` throw exceptions

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/CredentialIssuerMetaData.kt
```diff
@@ -301,3 +309,36 @@ value class PreferredKeyStorageStatusPeriod(
         require(value >= 31.days) { "Preferred key storage status period must be at least 31 days" }
     }
 }
+
+@JvmInline
+value class Wrprc private constructor(
+    val certificate: SignedJWT,
```
babisRoutis: Consider renaming the `certificate` to `value`
The name of the value class, (`Wrpc`) carries the schemantics.

src/main/kotlin/eu/europa/ec/eudi/pidissuer/domain/CredentialIssuerMetaData.kt
atsant: `orNull` has been removed and since `orThrow` is now the same with `tryParse`, `tryParse` has been made public and `orThrow` has been removed.

Moreover, we no longer return a `Result<WRPRC>` but a plain `WRPRC` object.

dzarras (approved):

